PR #1, opened 2016-06-29
ToCode Solution: תרגול תחביר

ynonp-test:


05_syntax_lab/01.py
```diff
@@ -1,5 +1 @@
-"""
-Write a program that reads 10 numbers from
-the user and prints the largest one
-"""
-
+solution for lab 1
```
ynonp, line 1: Very good solution.
Next time be sure to add comments and you can improve by doing proper error handling

05_syntax_lab/02.py
```diff
@@ -1,6 +1 @@
-"""
-Write a Python program that randomizes 7 numbers
-and prints their sum total.
-If the sum is divisable by 7, also print the word "Boom"
-"""
-
+my solution
```
szabgab, line 1: comment after pressing + on the last line of the 02.py examples.
Please keep the description of the exercises at the top of the file. It helps us remember what was the exercises about, and it will help you too. Oh and it is also a good practice to have a description of your program at the top of file.

05_syntax_lab/01.py
szabgab, line 1: Another comment for file 01.py

PR #2, opened 2016-07-03
ToCode Solution: תרגול תחביר

dovron:


05_syntax_lab/01.py
```diff
@@ -1,5 +1,16 @@
...
-
+"""
+      _      _      _
+   __(.)< __(.)> __(.)=
+   \___)  \___)  \___)   
+
+Python course - Verint
+Week 1
+
+"""
```
ynonp, line 9: Hi Dovron,
Love the use of string.format and the early submission :thumbsup:

Starting with big=0 is not the best choice here because users expect the biggest number to always be from their typed in values (and if all values are negative it won't work). Better to read the first value before the loop and read 9 more inside.

Also some alternatives to the if command worth knowing:
big = max(big, num)
Or python's ternary operator:
big = num if num > big else big

05_syntax_lab/02.py
```diff
@@ -1,6 +1,22 @@
...
+   __(.)< __(.)> __(.)=
+   \___)  \___)  \___)   
+
+Python course - Verint
+Week 1
+
+"""
+from random import randint
+
+random=randint(1, 100)
```
ynonp, line 12: Note unlike the previous exercise- here there's no reason to calculate the first value outside the loop. Also since loop index is not important here you can use the more concise:

```
sum = 0
for i in range(7): sum += randint(1,100)
```

05_syntax_lab/03.py
```diff
@@ -1,7 +1,19 @@
...
+"""
+      _      _      _
+   __(.)< __(.)> __(.)=
+   \___)  \___)  \___)   
+
+Python course - Verint
+Week 1
+
+"""
+from random import randint
```
ynonp, line 10: 👍 
1. do leave some spaces for readability for example `sum += random % 10`
2. it'd be nice to print the number before the loop so we can see the calculation worked

05_syntax_lab/04.py
```diff
@@ -1,8 +1,16 @@
...
+Week 1
+
+"""
+reverse=""
+while True:
+    line = raw_input()
+    if len(line) == 0: break
+    reverse = line + reverse
+
+print(reverse)
```
ynonp, line 16: Great solution (and +1 for the ducks)

print statement in python2 is usually written without the parenthesis (unlike python3 which forces parenthesis in print). The reason is printing multiple values separated by comma does not work as expected when using parenthesis

05_syntax_lab/05.py
```diff
@@ -1,5 +1,18 @@
...
+
+Python course - Verint
+Week 1
+
+"""
+from random import randint
+
+while True:
+    random=randint(1, 1000000)
+    if (random % 7) == 0:
```
ynonp, line 14: why not use `and` ?

05_syntax_lab/06.py
```diff
@@ -1,9 +1,21 @@
...
+Week 1
+
+"""
+from random import randint
+random1=randint(1, 10)
+random2=randint(1, 10)
+num = max(random1,random2)
+    
+while True:
+
```
ynonp, line 16: if X and Y is probably cleaner

05_syntax_lab/07.py
```diff
@@ -1,8 +1,27 @@
...
+    if (num > random):
+        if (fake % 10) == 0:
+            print "Try to multiple by ?"
+        else:
+             print "Too big"
+    elif (num < random):
+        print "Too small"
+    else: break
+
+print "Bingo!"
```
ynonp, line 27: Keep in mind that it's now easy to guess when computer is cheating: Once every 10 rounds and only if the guess happens to be larger than number.

If you'll have some extra time after finishing the modules exercise, perhaps you can go back to this one and make it harder for a user to know if the game's cheating

PR #3, opened 2016-07-04
ToCode Solution: תרגול תחביר

omerblecher:


szabgab:
Hi Omer,
thanks for the quick submission. Let me give you some input on exercise 01:
- `max` is a keyword in Python. It would be better to use a different variable name.
- The `print "Number {}:".format(n)` includes a newline at the end. Could you find a way to get rid of that newlines?
- Could you make some change so you won't need to ask the user for the first number separately, outside of the loop?

omerblecher:
Gabor Hi

Thanks for your comments. All comments applied.

Thanks
Omer

05_syntax_lab/01.py
```diff
@@ -1,17 +1,17 @@
 #Max Number
+import sys
```
szabgab: Nice. Now try without using `sys`.
- You don't need to set a maximum outside the loop
- `raw_input` can have a parameter

05_syntax_lab/02.py
```diff
@@ -1,6 +1,15 @@
...
+
+#Get 7 numbers
+for number in range(7):
+    sum += randint(1, 100)
+
+print "The sum of the 7 random numbers is {}".format(sum)
+
+# check if the sum is divided by 7
+if sum % 7 == 0:
+    print "boom"
```
szabgab, line 15: Looks good. 👍

05_syntax_lab/03.py
```diff
@@ -1,7 +1,16 @@
...
+print "The number is: {}".format(number)
+
+sum = 0
+
+#Loop over the digits
+while number > 0:
+    sum += number % 10
+    number /=10
+
+print "The sum of all digits is {}".format(sum)
```
szabgab, line 16: Looks good.

05_syntax_lab/04.py
```diff
@@ -1,8 +1,17 @@
...
+    userInput = raw_input()
+    if len(userInput) == 0:
+        break
+    results = userInput + "\n" + results
+
+#print all lines backward
+if len(results) > 0:
+    print results
+else:
+    print "Empty input"
```
szabgab, line 21: That looks ok. Could you also implement a version where you store the individual input lines separately?

05_syntax_lab/01.py
omerblecher: Done

05_syntax_lab/04.py
omerblecher, line 21: Done

05_syntax_lab/01.py
```diff
@@ -3,7 +3,6 @@
 print "Enter 10 numbers: "
 print " "
 
-maxNumber = 0
 
 # loop for the next 10 numbers
 for n in range(1, 11):
```
szabgab, line 8: 👍

PR #4, opened 2016-07-04
ToCode Solution: תרגול תחביר

Guyemin:


05_syntax_lab/01.py
```diff
@@ -1,5 +1,8 @@
-"""
-Write a program that reads 10 numbers from
-the user and prints the largest one
-"""
-
+print "Please write 10 numbers"
```
szabgab, line 1: Thanks for the first solution. A few comments inline.

05_syntax_lab/01.py
```diff
@@ -1,5 +1,8 @@
-"""
-Write a program that reads 10 numbers from
-the user and prints the largest one
-"""
-
+print "Please write 10 numbers"
+
+high = int(raw_input())
```
szabgab, line 3: Could you eliminate the duplication and call `raw_input` only inside the loop?

05_syntax_lab/01.py
```diff
@@ -1,5 +1,8 @@
-"""
-Write a program that reads 10 numbers from
-the user and prints the largest one
-"""
-
+print "Please write 10 numbers"
+
+high = int(raw_input())
+for i in range (1,10):
+	number = int(raw_input())
```
szabgab, line 5: Could you include a prompt message that will tell the user to type in the i-th number?

05_syntax_lab/02.py
```diff
@@ -1,6 +1,9 @@
...
-
+from random import randint
+sum = 0
+for i in range (1,8):
+	x = randint(0,100)
+	print x
+	sum = sum + x
+print "The sum is %s" % sum
+if sum % 7 == 0:
+	print "Boom"
```
szabgab, line 9: Looks good.

05_syntax_lab/03.py
```diff
@@ -1,7 +1,11 @@
...
+sum = 0
+x = randint(1,10000)
+print "The number is %s" % x
+while (x % 10) > 0:
+	num = x % 10
+	print num
+	sum = sum + num
+	x = x / 10
+	
+print "The sum is: %s" % sum
```
szabgab, line 11: Looks good. 👍

05_syntax_lab/04.py
```diff
@@ -1,8 +1,11 @@
...
+word = raw_input()
+total = word
+index = 0
+while word != ' ':
+	if index > 0:
+		total = "%s %s"  % (word, total)
+	print "Please add another word"
+	word = raw_input()
+	index = index + 1
+print total
```
szabgab, line 11: Please simplify this solution eliminating the two separate calls to `raw_input` and eliminating the `index`.
BTW these are rows you read in not words and thus you need to add newline between them and not space.
You could also use `+` to concatenate two strings.

05_syntax_lab/05.py
```diff
@@ -1,5 +1,14 @@
...
+	index = index + 1
+	if x % 7 == 0:
+		if x % 13 == 0:
+			if x % 15 == 0:
+				print "Bingo!"
+				print "This is the number: %s" % x
+				index = 0
+	x = randint(1,1000000)
+	
+
```
szabgab, line 14: - Why do you need the `index` here?
- The 3 conditions create a strange looking steps. In such cases usually it is better to have a single `if` conditions. Can you do that?
- Could you eliminate the duplicate call to `randint`?

05_syntax_lab/07.py
```diff
@@ -1,8 +1,17 @@
...
+	if x > guess: 
+		print "Your number is lower"
+		index = index + 1
+	elif x < guess: 
+		print "You number is higher"
+		index = index + 1
+	else:
+		print "Bingo!"
+		index == 0
+		break
```
szabgab, line 17: Here too, I don't see why do you need the `index` variable.

05_syntax_lab/06.py
```diff
@@ -1,9 +1,13 @@
...
+y = int(randint(1,10))
+print "The X number is %s" % x
+print "The Y number is %s" % y
+while index != 0:
+	if (index % x == 0):
+		if (index % y == 0):
+			print index 
+			index = 0
+			break
+	index = index + 1
```
szabgab, line 13: Hi,
`index` will only become 0 in line 11, but you then call `break` anyway, so the comparison in the `while` statement will never be false. It is better to state that explicitly by writing `True` there.
The solution while correct will be rather slow if the numbers are big. Could you find a better algorithm and implement that?

PR #5, opened 2016-07-04
ToCode Solution: תרגול תחביר

eladagam:


05_syntax_lab/01.py
```diff
@@ -1,5 +1,18 @@
...
+Max = input
+
+for num in range(9):
+	print "Enter another Number"
+	input = raw_input()
+	if input > Max:
+		Max=input
+
+print "Max = {maximum}".format(maximum=Max)
+
```
szabgab, line 17: Hi Elad,
thanks for your first solution.
Your solution seems to work, but there can be a number of improvements. Please try them:
- Variable names are usually all lowercase, but `max` is also a keyword in Python. Could you use a different variable name?
- Could you eliminate the duplicate call to `raw_input`?
- The call to `format` with the name is nice very nice!
- The "Enter another Number" now prints a newline before the user can enter the input. Could you get rid of that newlines?

05_syntax_lab/02.py
```diff
@@ -1,6 +1,19 @@
...
+sum = 0 
+
+for num in range(7):
+	sum = sum + randint(1, 100)
+
+print "Sum = {Summary}".format(Summary=sum)
+if sum % 7 == 0:
+	print "Boom"
+
+
```
szabgab, line 19: Nice! 👍

05_syntax_lab/03.py
```diff
@@ -1,7 +1,19 @@
...
+
+rand = randint(1, 10000)
+print "Rand = {R}".format(R=rand)
+for num in str(rand):
+	sum = sum + int(num)
+
+print "Sum = {Summary}".format(Summary=sum)
+
+
+
```
szabgab, line 19: Extra point for using `for` loop on a string! 🍪

05_syntax_lab/04.py
```diff
@@ -1,8 +1,22 @@
...
+	else:
+		break
+
+
+numoflines =(len(Input)) 
+for i in range(numoflines):
+	print Input.pop()
+
+
+
```
szabgab, line 22: Nice! Especially the use of lists.
It would be better to use all-lower-case variable names but not `input` as that is a reserved word. Maybe `line` ?
It  might be better to print the lines without destroying the Input. Can you do that?

05_syntax_lab/05.py
```diff
@@ -1,5 +1,15 @@
...
+from random import randint
+
+ 
+while True:
+	num = randint(1, 1000000)
+	if num % 7 == 0 and num % 13 == 0 and num % 15 == 0:
+		print "Found a number that divides to 7 13 15 => {n}".format(n=num)
+		break
+
+
```
szabgab, line 15: 👍

05_syntax_lab/04.py
eladagam, line 22: Done

05_syntax_lab/01.py
eladagam, line 17: Done

PR #6, opened 2016-07-04
ToCode Solution: תרגול תחביר

lahavz:


05_syntax_lab/01.py
```diff
@@ -1,5 +1,11 @@
...
+
+num = 0
+for i in range(10):
+    print "Number # %d:" % (i+1)
+    tmpnum = int(raw_input())
+    if tmpnum >= num:
+        num = tmpnum
+
+
+print "Highest Number: %d" % num
```
ynonp, line 11: 1. Better to leave the comment line at the top. That way when you open the script next week you'll still remember what it does 😄 
2. What would be the result if all numbers are negative? Can you make it less surprising?
3. I think prompting for the number would be better without the newline character (in line 5). Can you find a way to print and let the user type in the input in the same line?

05_syntax_lab/02.py
```diff
@@ -1,6 +1,12 @@
...
+print "Exercise # 2"
+
+
+for i in range(7):
+    num = randint(1, 100)
+    print "Number # %d is %d" % ((i+1),num)
+    if num % 7 == 0:
+        print "Booooommmmm"
+        
+
```
ynonp, line 12: Nice but you're not calculating the sum... (note the exercise)

05_syntax_lab/03.py
```diff
@@ -1,7 +1,15 @@
...
-The result should be: 14
-"""
-
+from random import randint
+
+print "Exercise # 3"
+
+num = randint(1, 10000)
+print "Number is %d" % (num)
+sum = int(0)
```
ynonp, line 7: no need for int(0). `sum = 0` is the same and more readable

05_syntax_lab/04.py
```diff
@@ -1,8 +1,16 @@
...
+while True:  
+    line = raw_input()
+    if len(line) <= int(0): break
+    array.append(line)       
+
+
+
+for i in range((len(array)-1), -1 , -1):
+    print array[i]
+
```
ynonp, line 16: Nice! Can you also find a way to solve this without using arrays?

05_syntax_lab/05.py
```diff
@@ -1,5 +1,22 @@
...
+                break
+            else:
+                print "Number %d is NOT divided by 15" % (num)
+        else:
+                print "Number %d is NOT divided by 13" % (num)
+    else:
+                print "Number %d is NOT divided by 7" % (num)
+                
+        
+
```
ynonp, line 37: Can you find a way to write the same code using less indentation? (look how the if/else keywords make a little triangle, which makes it harder to understand what belongs to what). 
Hint: Search for `continue` keyword

05_syntax_lab/06.py
```diff
@@ -1,9 +1,20 @@
...
+
+while True:
+    if ((i*num1) % num2) == 0:
+        print "Smallest multiply is %d" % (i*num1)
+        break
+    else:
+        i = i+1
+        
+
+
```
ynonp, line 20: Nice 👍

05_syntax_lab/07.py
```diff
@@ -1,8 +1,37 @@
...
+            print "Try higher number"
+        else:
+            print "Try smaller number"
+        i = int(-1)
+
+    i=i+1
+    
+    
+            
+
```
ynonp, line 37: Well Done!

I see you're very close to finishing the second lab too, so as a bonus you can try reducing some repetitions in this number guessing game. It may require concepts we haven't learned yet so feel free to explore

05_syntax_lab/05.py
lahavz, line 37: I can make it within the same if statement using and :)
if ((num % 7 == 0) && (num % 13 == 0)  && (num % 15 == 0)):

or using elif :

while True:
    num = randint(1, 1000000)
    print "Number is %d" % (num)
    if num % 7 > 0:
        continue
    elif num % 13 > 0:
        continue
    elif num % 15 > 0:
        continue
    else:
        print "Number %d is divided by 7,13,15" % (num)
        break

05_syntax_lab/05.py
```diff
@@ -17,6 +17,21 @@
                 print "Number %d is NOT divided by 13" % (num)
     else:
                 print "Number %d is NOT divided by 7" % (num)
+"""
+
+#New Block (fix with less indentation)
```
ynonp, line 22: We're using github :) No need to leave original block. Github shows you the diff automatically and keeps track of every past submission

05_syntax_lab/05.py
```diff
@@ -17,6 +17,21 @@
                 print "Number %d is NOT divided by 13" % (num)
     else:
                 print "Number %d is NOT divided by 7" % (num)
+"""
+
+#New Block (fix with less indentation)
+while True:
+    num = randint(1, 1000000)
+    print "Number is %d" % (num)
+    if num % 7 > 0:
```
ynonp, line 26: Cool, also consider `if num % 7 > 0: continue` (in the same line)

05_syntax_lab/05.py
```diff
@@ -17,6 +17,21 @@
...
+while True:
+    num = randint(1, 1000000)
+    print "Number is %d" % (num)
+    if num % 7 > 0:
+        continue
+    elif num % 13 > 0:
+        continue
+    elif num % 15 > 0:
+        continue
+    else:
```
ynonp, line 32: No need for the else here. The only way to reach this point is to have all other conditions false (since each has a continue)

PR #7, opened 2016-07-04

lahavz:


06_modules_lab/01.py
```diff
@@ -1,4 +1,6 @@
-""" Write a program that takes a count
-from sys.argv import and prints "Hello Python"
-count times
-"""
+import sys
+
+
+for i in range(int(sys.argv[1])):
+    print "Hello World"
+
```
ynonp, line 6: 👍

06_modules_lab/02_bonus.py
```diff
@@ -1,10 +1,11 @@
...
+
+num1 = int(0)
+num2 = int(0)
+
+try:
+    num1 = int(sys.argv[1])
+    num2 = int(sys.argv[2])
+    print "Sum is %d" % (num1+num2)
+except:
+    print "Problematic Agruments"
```
ynonp, line 11: Fantastic! But do be kind and tell the user how to use your script

06_modules_lab/03_bonus.py
```diff
@@ -1,13 +1,32 @@
...
+
+
+for root, dirs, files in os.walk(sys.argv[1]):
+#for root, dirs, files in os.walk("C:\Users\lzeno\OneDrive\Code\Python Scripts\ToCode Course\Tasks\Models"):
+    for name in files:
+        if os.path.getsize(os.path.join(root,name)) >= int(sys.argv[2]):
+            print os.path.join(root,name) + " is bigger than " + sys.argv[2] + "Bytes Delete? (Y/N)"
+            answer = raw_input()
+            while True:
+                if answer == 'Y' or answer == 'y':
```
ynonp, line 20: I'd use `answer.lower() == 'y'` instead

06_modules_lab/03_bonus.py
```diff
@@ -1,13 +1,32 @@
...
+
+args = parser.parse_args()
+
+
+
+for root, dirs, files in os.walk(sys.argv[1]):
+#for root, dirs, files in os.walk("C:\Users\lzeno\OneDrive\Code\Python Scripts\ToCode Course\Tasks\Models"):
+    for name in files:
+        if os.path.getsize(os.path.join(root,name)) >= int(sys.argv[2]):
+            print os.path.join(root,name) + " is bigger than " + sys.argv[2] + "Bytes Delete? (Y/N)"
```
ynonp, line 17: 1. Can you use str.format to make this line more readable? Can you remove the newline at the end?
2. Usually yes/no question uses one of the letters as uppercase to suggest a default. So you can use "Delete ? (y/N)" and if the user types anything other than a "y" file is not deleted

PR #8, opened 2016-07-04
ToCode Solution: תרגול תחביר

konst1988:


05_syntax_lab/01.py
```diff
@@ -1,5 +1,11 @@
-"""
-Write a program that reads 10 numbers from
-the user and prints the largest one
-"""
-
+#"""
```
ynonp: You don't need both # and """. Usually we use """ at the beginning of the file for general file description, and # for line comments throughout the script

05_syntax_lab/01.py
```diff
@@ -1,5 +1,11 @@
...
+#Write a program that reads 10 numbers from
+#the user and prints the largest one
+#"""
+print "Enter 10 numbers:"
+max = 0
+for i in range(0,10):
```
ynonp: 1. Consider the case if all 10 numbers a user enters are negatives. What will the program print? How can you fix it?
2. Prefer using `range(10)` over `range(0,10)` - it's shorter and cleaner

05_syntax_lab/01.py
```diff
@@ -1,5 +1,11 @@
...
+    num = int(raw_input("Enter number[{}]:".format(i+1)))
```
ynonp, line 9: 👍

05_syntax_lab/02.py
```diff
@@ -1,6 +1,17 @@
...
 
+from random import randint
+sum = 0
+for i in range(0,10):
+    num = randint(1,100)
+    print "generated num:" ,num
+    sum+=num
+    
+
+print "sum is " ,sum
```
ynonp: Oh no too many spaces :) print already writes a space between its arguments so no need for the space inside quotes. A less confusing alternative BTW is to use str.format or %

05_syntax_lab/02.py
```diff
@@ -1,6 +1,17 @@
...
+if sum % 7  == 0:
+        print "BOOM!"
```
ynonp, line 17: 💣 💥

05_syntax_lab/01.py
```diff
@@ -1,10 +1,11 @@
...
+"""
+
 print "Enter 10 numbers:"
-max = 0
-for i in range(0,10):
+max = int(raw_input("Enter number[1]:"))
+for i in range(1,9):
     num = int(raw_input("Enter number[{}]:".format(i+1)))
     if num > max:
         max = num
```
ynonp, line 11: Nice!

05_syntax_lab/02.py
```diff
@@ -12,6 +12,6 @@
     sum+=num
     
 
-print "sum is " ,sum
+print "sum is {}".format(sum)
```
ynonp, line 15: 👍

05_syntax_lab/03.py
```diff
@@ -5,3 +5,12 @@
 The result should be: 14
 """
 
+from random import randint
+num = randint(1,10000)
+print "num is:{}".format(num)
+sum = 0
+while num > 0:
+    digit = num % 10
+    sum += digit
```
ynonp, line 14: Notice how github colors "sum" and "print" differently than "num". The reason is sum() is a function in python. Using it as a variable name works but can be confusing. `sum_digits` or `result` would be less confusing

05_syntax_lab/05.py
```diff
@@ -3,3 +3,10 @@
 until it finds a number that is divisable by: 7, 13 and 15
 """
 
+from random import randint
+
+while True:
+    num = randint(1,1000000)
+    if num % 7 == 0 and num % 13 and num % 15 == 0:
+        print num
+        break
```
ynonp, line 12: 👍

PR #10, opened 2016-07-04
ToCode Solution: תרגול תחביר

eladkah:


ynonp:
Hi Elad and Welcome to the course,

In the future, please, keep the comments at the beginning of the file to make it easier
for use to know what is the exercise about. It will also help you when you look back at your code a few week from now.

See other file specific feedback inline

05_syntax_lab/01.py
```diff
@@ -1,5 +1,7 @@
-"""
-Write a program that reads 10 numbers from
-the user and prints the largest one
-"""
-
+# First number is biggest
+max = int(raw_input())
```
ynonp, line 2: Can you add a nice prompt so the user will know what they're typing?

05_syntax_lab/02.py
```diff
@@ -1,6 +1,8 @@
...
-"""
-
+from random import randint
+
+sum = 0
+for i in range(0,7):
+    sum += randint(1,100)
+print 'Sum of 7 random numbers is:', sum
+if (sum % 7 == 0):
+    print 'Boom'
```
ynonp, line 8: No need for parenthesis in python if + one line if are usually better written as:
`if sum % 7 == 0: print 'Boom!'`

05_syntax_lab/03.py
```diff
@@ -1,7 +1,9 @@
...
-
+from random import randint
+
+sum = 0
+rand = randint(1,10000)
+original = rand
+while (rand > 0):
+    sum += rand % 10
+    rand /= 10
+print 'Rand:', original, " sum of digits:", sum
```
ynonp, line 9: Can you use str.format or % to make this line more readable?

05_syntax_lab/04.py
```diff
@@ -1,8 +1,10 @@
...
+
+line = "dummy"
+arr = []
+index = 0
+while (line != ""):
+    line = raw_input()
+    arr.insert(index, line)
+    index+=1
+for i in range(1,index + 1):
+    print arr[index-i]
```
ynonp, line 10: 1. Nice! Can you also find a way to solve this without using arrays?
2. While we haven't yet learned arrays, there's  a simpler way to print the reversed list:

```
for item in reversed(arr): print item
```

05_syntax_lab/05.py
```diff
@@ -1,5 +1,7 @@
...
-until it finds a number that is divisable by: 7, 13 and 15
-"""
-
+from random import randint
+
+while True:
+    rand = randint(1,1000000)
+    if rand % 7 == 0 and rand % 13 == 0 and rand % 15 == 0:
+        print rand
+        break
```
ynonp, line 7: 👍

05_syntax_lab/06.py
```diff
@@ -1,9 +1,11 @@
...
+import sys
+
+x = randint(1,10)
+y = randint(1,10)
+print 'x: ', x, 'y: ', y
+for ix in range(1, x+1):
+    for iy in range(1, y+1):
+        if y * ix == x * iy:
+            print y * ix
+            sys.exit(0)
```
ynonp, line 11: Python has a gcd function built-in (read here: https://docs.python.org/2/library/fractions.html#fractions.gcd)

Can you use that to simplify your solution?

05_syntax_lab/07.py
```diff
@@ -1,8 +1,21 @@
...
+
+guess = 0
+while True:
+    guess = int(raw_input())
+    if guess == rand:
+        print 'You are right!'
+        break
+    # randomize lie
+    lie = randint(0,2)
+    if lie == 0:
```
ynonp, line 13: Python also supports the shorter:
`large = guess < rand if lie == 0 else guess > rand`

Though I'm not sure which is more readable

PR #12, opened 2016-07-05
ToCode Solution: תרגול תחביר

zfigov:
Updated files

### DIFF
--- a/05_syntax_lab/01.py
+++ b/05_syntax_lab/01.py
@@ -2,4 +2,14 @@
 Write a program that reads 10 numbers from
 the user and prints the largest one
 """
+# Exercise 1
+# By Zvi Figov
+print "Enter 10 positive numbers" # Assuming positive numbers 
+largest = -1
+for i in range(1,11):
+    print "Enter number %d" % i
+    num = int(raw_input())
+    largest = max(num,largest)
+
+print "Largest number is %d" % largest
 
--- a/05_syntax_lab/01.py
+++ b/05_syntax_lab/01.py
@@ -4,10 +4,13 @@ the user and prints the largest one
 """
 # Exercise 1
 # By Zvi Figov
-print "Enter 10 positive numbers" # Assuming positive numbers 
+import sys
+
+print "Enter 10 numbers" 
 largest = -1
 for i in range(1,11):
-    print "Enter number %d" % i
+    txtstr =  "Enter number %d : " % i
+    sys.stdout.write(txtstr)
     num = int(raw_input())
     largest = max(num,largest)
 
--- a/05_syntax_lab/02.py
+++ b/05_syntax_lab/02.py
@@ -3,4 +3,17 @@ Write a Python program that randomizes 7 numbers
 and prints their sum total.
 If the sum is divisable by 7, also print the word "Boom"
 """
+# Exercise 2
+# By Zvi Figov
+from random import randint
+
+acc = 0
+# accumulate numbers from user
+for i in range(1,8):
+    x = randint(1,100)
+    print "Number %d is %d" % (i, x)
+    acc = acc + x
+print "Accumulated sum = %d" % acc
+# test if they divide by 7
+if acc % 7 ==0 : print "Boom!"
 
--- a/05_syntax_lab/03.py
+++ b/05_syntax_lab/03.py
@@ -11,7 +11,8 @@ from random import randint
 num = randint(1,10000)
 numstr = str(num)
 acc = 0
-for n in range(0,len(numstr)):
-    acc = acc + int(numstr[n])
+for n in numstr:
+    acc = acc + int(n)
 print "The sum of the digits of the number %d is %d" % (num,acc)
+
 
--- a/05_syntax_lab/03.py
+++ b/05_syntax_lab/03.py
@@ -4,4 +4,14 @@ and prints the sum total of its digits.
 For example if the number was: 2345
 The result should be: 14
 """
+# Exercise 3
+# By Zvi Figov
+from random import randint
+
+num = randint(1,10000)
+numstr = str(num)
+acc = 0
+for n in range(0,len(numstr)):
+    acc = acc + int(numstr[n])
+print "The sum of the digits of the number %d is %d" % (num,acc)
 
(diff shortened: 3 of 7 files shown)


05_syntax_lab/01.py
```diff
@@ -2,4 +2,14 @@
...
+# Exercise 1
+# By Zvi Figov
+print "Enter 10 positive numbers" # Assuming positive numbers 
+largest = -1
+for i in range(1,11):
+    print "Enter number %d" % i
+    num = int(raw_input())
+    largest = max(num,largest)
+
+print "Largest number is %d" % largest
```
szabgab, line 17: Welcome to the exercises!

Your solution works in some of the cases, but what if all the numbers are less than -10? Could you fix the code so there won't be such limitation?

In addition, when you prompt the user to enter a number, the cursor is on the line following the prompt. Could you change that so the prompt will be on the same line as the "Enter number 2" text ?

05_syntax_lab/02.py
```diff
@@ -3,4 +3,17 @@
...
+
+acc = 0
+# accumulate numbers from user
+for i in range(1,8):
+    x = randint(1,100)
+    print "Number %d is %d" % (i, x)
+    acc = acc + x
+print "Accumulated sum = %d" % acc
+# test if they divide by 7
+if acc % 7 ==0 : print "Boom!"
```
szabgab, line 18: 👍

05_syntax_lab/03.py
```diff
@@ -4,4 +4,14 @@
...
+# Exercise 3
+# By Zvi Figov
+from random import randint
+
+num = randint(1,10000)
+numstr = str(num)
+acc = 0
+for n in range(0,len(numstr)):
+    acc = acc + int(numstr[n])
+print "The sum of the digits of the number %d is %d" % (num,acc)
```
szabgab, line 17: You can iterate over the characters of a string using `for c in numstr:` that will further simplify the code.

05_syntax_lab/04.py
```diff
@@ -5,4 +5,19 @@
...
+total_lines = ""
+print "Enter text, empty line to stop"
+
+line = raw_input()
+while len(line)>0:
+    print "Input line %s" % line
+    total_lines = line + "\n" + total_lines
+    line = raw_input()
+
+print "Lines reversed: \n%s" % total_lines
```
szabgab, line 23: I think it would be better to call  `raw_input()` only once. Could you do that?

05_syntax_lab/05.py
```diff
@@ -2,4 +2,15 @@
...
+# By Zvi Figov
+from random import randint
+
+flag_done = False
+
+while not flag_done:
+    num = randint(1,1000000)
+    if num % 7 ==0 and num % 13 ==0 and num % 15 ==0:
+        flag_done = True
+print "Chosen number = %d" % num
```
szabgab, line 15: Try to use `break` to exit an infinite `while` loop. That way you don't need to use the extra variable `flag_done`

05_syntax_lab/06.py
```diff
@@ -6,4 +6,35 @@
...
+            common_mult = common_mult + i + "\n"
+# get minimum common mult
+min_c = 999999
+for i in common_mult.split():
+    if int(i)<min_c :
+        min_c = int(i)
+print "Least common multiplier of %d and %d = %d" % (num1,num2,min_c)
+
+
+
```
szabgab, line 15: Using strings when we are dealing with numbers is quite confusing. Please don't do that.
I am not sure if generating all the multipliers of both numbers is efficient, but I think there are better solutions. Try to look for a faster algorithm. Especially think about the case when you have two large numbers to work with.

05_syntax_lab/07.py
```diff
@@ -5,4 +5,20 @@
...
+while not found:
+    print "Enter guess" 
+    guess_num = int(raw_input())
+    if num==guess_num:
+        print "Well done"
+        found = True
+    elif guess_num > num:
+        print "Too big"
+    else:
+        print "Too small"
```
szabgab, line 24: 👍

05_syntax_lab/01.py
zfigov, line 17: I added the limitation in the request from the user (so it works in the cases I defined) but I can change that.
Same prompt - I found write does that.

05_syntax_lab/01.py
```diff
@@ -4,10 +4,13 @@
...
+import sys
+
+print "Enter 10 numbers" 
 largest = -1
 for i in range(1,11):
-    print "Enter number %d" % i
+    txtstr =  "Enter number %d : " % i
+    sys.stdout.write(txtstr)
     num = int(raw_input())
     largest = max(num,largest)
```
szabgab, line 16: I guess I managed to mislead you. Sorry for that.

Yours is a solution but you actually don't need `sys`, not even `print` to show your prompt. Read how `raw_input` works. https://docs.python.org/2/library/functions.html#raw_input

Also, please replace the older `%` with the newer `.format` for creating strings with values in it.

05_syntax_lab/03.py
```diff
@@ -11,7 +11,8 @@
 num = randint(1,10000)
 numstr = str(num)
 acc = 0
-for n in range(0,len(numstr)):
-    acc = acc + int(numstr[n])
+for n in numstr:
+    acc = acc + int(n)
 print "The sum of the digits of the number %d is %d" % (num,acc)
 
+
```
szabgab, line 18: 👍

05_syntax_lab/05.py
```diff
@@ -6,11 +6,10 @@
...
 
-flag_done = False
 
-while not flag_done:
+while True:
     num = randint(1,1000000)
     if num % 7 ==0 and num % 13 ==0 and num % 15 ==0:
-        flag_done = True
+        break
 print "Chosen number = %d" % num
```
szabgab, line 15: 👍

05_syntax_lab/06.py
```diff
@@ -12,29 +12,9 @@
...
-print "Least common multiplier of %d and %d = %d" % (num1,num2,min_c)
-
-
-
 
+m1 = min(num1,num2)
+for i in range(m1,num1*num2+1,m1):
+    if i % num1 ==0 and i % num2 ==0:
+        print "Least common multiplier of %d and %d = %d" % (num1,num2,i)
+        break
```
szabgab, line 20: That's much better code, but the algorithm is still rather slow. Could you find a faster algorithm and implement that?

PR #14, opened 2016-07-05
ToCode Solution: תרגול תחביר

HananRosenthal:


05_syntax_lab/01.py
```diff
@@ -1,5 +1,7 @@
...
-the user and prints the largest one
-"""
-
+############### Part 1
+print('Please enter 10 numbers in one raw, separated by a space')
+numbers = input()
+print('These are the numbers:')
+print(numbers)
+print('The biggest value is:')
+print(max(numbers))
```
szabgab, line 7: Using `input()` in Python 2 is dangerous. Please use `raw_input` instead of it.

05_syntax_lab/02.py
```diff
@@ -1,6 +1,12 @@
...
+L = 7
+for i in range(0,L):
+    r = randint(1,100)
+    sum += r
+    print('%d' % r)
+
+print('The sum of the seven random numbers is {}'.format(sum))
+if sum % L == 0:
+    print('%s' % 'Boom')
+#How many times (on the average) should we run this program in order to get a 'Boom'?
```
szabgab, line 12: 👍

05_syntax_lab/01.py
HananRosenthal, line 7: I am working with python 3.3.2 (see “print” and of course “input”).
This is the installation on my PC.
10X

szabgab, line 7: OK, I understand from @ynonp it is ok for you to use Python 3. I'll check your solutions according.

In that case `input` is ok, though I think for the sake of this exercises it would be better to expect each number on a separate line.

05_syntax_lab/03.py
```diff
@@ -1,7 +1,12 @@
...
+r = randint(1, L)
+str_r = str(r)
+print('The string number is {}'.format(str_r) )
+num_digits = len(str_r)
+print('The number %d contains %d digits' % (r,num_digits) )
+sum = 0
+for i in range(0, num_digits):
+    sum += int(str_r[i])
+
+print('The digits sum is {}'.format(sum) )
```
szabgab, line 12: - You don't need an index to iterate over characters of a string you can write `for c in str_r`
- It would be better to consistently use `.format` instead of the old `%` syntax.

05_syntax_lab/04.py
```diff
@@ -1,8 +1,14 @@
...
+    tmp = input()
+    if len(tmp) <= 0:
+        break
+    text.append(tmp)
+    #print(text)
+
+i = len(text)
+while i > 0:
+    print(text[i-1])
+    i -= 1
```
szabgab, line 14: - It is better to use `True` instead of 1 in the infinite while loop. That's the real `True` of Python.
- `input()` can have a parameter which is the prompt text so you don't need to print separately.

05_syntax_lab/05.py
```diff
@@ -1,5 +1,7 @@
...
-until it finds a number that is divisable by: 7, 13 and 15
-"""
-
+####################### Part 5
+while 1:
+    r = randint(1, 1000000)
+    if (r % 7 == 0) and (r % 13 == 0) and (r % 15 == 0):
+        print('{}'.format(r))
+        break
+
```
szabgab, line 7: - `True` instead 1 here too, otherwise looks good. 👍

05_syntax_lab/06.py
```diff
@@ -1,9 +1,15 @@
...
+i = 1
+min_quotient = []
+while i <= maximum:
+    if (i % r1 == 0) and (i % r2 == 0):
+        min_quotient.append(i)
+        #print('{}'.format(min_quotient))
+    i += 1
+
+print('%d' % min(min_quotient))
+
```
szabgab, line 15: This algorithm will be quite slow if the numbers are big (not just 10*9). Could you find and implement a faster one?

05_syntax_lab/07.py
```diff
@@ -1,8 +1,15 @@
...
+    r = int(input())
+    if r == r1:
+        print('BINGO')
+        break
+    if r < r1:
+        print('your guess ({}) is too small'.format(r))
+    else:
+        print('your guess ({}) is too big'.format(r))
+
+########################### End of Syntax exercise ############################
```
szabgab, line 15: Please use `while True` for the infinite loop. Also could you please use more descriptive variable names?

05_syntax_lab/04.py
HananRosenthal, line 14: Hello Gabor,

Last comment is not clear, please elaborate (“input() can have a parameter which is the prompt text so you don't need to print separately”)
10X

Hanan Rosenthal
R&D Research&Develompment

Desk: +972.9.962.5985
Mobile: +972.54.778.5985
Verint Communications and Cyber Intelligence Solutions
www.verint.comhttp://www.verint.com/

szabgab, line 14: I mean this:
`v = input('prompt')`

05_syntax_lab/06.py
HananRosenthal, line 15: Maybe this one:

from random import randint

r1 = randint(1, 10)
r2 = randint(1, 10)

print('%d  %d' % (r1, r2))

mini = min(r1,r2)
maxi = max(r1,r2)
i = int(maxi/mini)

while True:
    if i*mini % maxi == 0:
        break
    i += 1

print('The least common multiple of the two numbers is: {}'.format(i*mini))

szabgab, line 15: Yes, that looks faster. 👍

05_syntax_lab/04.py
HananRosenthal, line 14: Hi,

I am using:

#-_\- coding: utf-8 -_-
print(u'\u05d1\u05e8\u05d5\u05db\u05d9\u05dd \u05d4\u05d1\u05d0\u05d9\u05dd')

but get gibberish…
how come?

PR #15, opened 2016-07-05

HananRosenthal:


06_modules_lab/01.py
```diff
@@ -1,4 +1,15 @@
...
+    print('Usage: %s <number of repetions>' % sys.argv[0])
+    sys.exit(1)
+
+(_, num) = sys.argv
+print('number of repetitions = {}'.format(num) )
+#print('{}'.format(type(num)))
+
+for i in range (0, int(num)):
+    print('Hello Python')
+
```
szabgab, line 15: 👍 
Minor issues:
- You don't need to import `os` here
- Better to always use `format` and not `%`.
- Better to write `num = sys.argv[1]`

06_modules_lab/02.py
```diff
@@ -1,4 +1,14 @@
...
+if len(sys.argv) != 3:
+    print('Usage: %s <num1>  <num2> ' % sys.argv[0])
+    sys.exit(1)
+
+(_, num1, num2) = sys.argv
+num1 = int(num1)
+num2 = int(num2)
+print('%d %d %d' % (num1, num2, num1+num2))
+
+
```
szabgab, line 16: 👍 
Same minor comments as on 01

06_modules_lab/03.py
```diff
@@ -1,11 +1,30 @@
...
+    #print('%s'% name)
+    #statinfo = os.stat(path +  name)
+    #print(statinfo)
+    size = os.path.getsize(args['path'] + name)
+    if size > int(args['minsize']):
+        #print(size)
+        print('Delete %s? [1 for YES 0 for NO]' % (args['path'] + name))
+        ans = input()
+        if int(ans) == 1:#delete the file
+           os.remove(args['path'] + name)
```
szabgab, line 30: I tried the program passing `-p . -ms 1` and got an exception `No such file or directory: '..git'`
BTW I'd not convert the `ans` to number, I'd just compare it to the string `'1'`

HananRosenthal, line 30: This is because you enter a ‘path’ not ending with ‘\’

HananRosenthal, line 30: This should do…..

import os
import sys
import argparse

parser = argparse.ArgumentParser(description='The program gets a path where file sizes over a pre-defined value are deleted, if required by the user')
parser.add_argument('-p','--path', help='full path (\ - terminated) to search', required=True)
parser.add_argument('-ms','--minsize', help='The size threshold', required=True)

kuku = parser.parse_args()

suffix = "\"
if (kuku.path).endswith(suffix) == False:
    kuku.path += suffix

for name in os.listdir(kuku.path):
    #print('%s'% name)
    #statinfo = os.stat(kuku.path +  name)
    #print(statinfo)
    size = os.path.getsize(kuku.path + name)
    if size > int(kuku.minsize):
        ans = input('Delete {}? 1=Yes 0=No '.format(kuku.path + name))
        if ans == '1':
           os.remove(kuku.path + name)

szabgab:
A technical question: In the previous two cases, have you responed to the e-mail with your solution or have you submitted it as you did with the original solutions?

HananRosenthal:
I have responded by e-mail

szabgab:
I see. That makes it very hard to evaluate your solution. From now on, please use the regular submission.

PR #18, opened 2016-07-05
ToCode Solution: תרגול תחביר

razavishay:


05_syntax_lab/01.py
```diff
@@ -1,5 +1,9 @@
-"""
-Write a program that reads 10 numbers from
-the user and prints the largest one
-"""
-
+"""
+receive 10 numbers and print the largest of them
+"""
+max = int(raw_input())
```
ynonp, line 4: Fantastic! Just one comment-
Notice how github colours "max" in the same color as print (and differently from "number" and "Boom"). The reason is python has a function named max, so when you use that as a variable name it can be a bit confusing.

05_syntax_lab/01.py
```diff
@@ -1,5 +1,9 @@
...
-
+"""
+receive 10 numbers and print the largest of them
+"""
+max = int(raw_input())
+for i in range (9):
+    b = int(raw_input())
+    if b > max:
+        max = b
+print "Largest number is:", max
```
ynonp, line 9: Consider using str.format when combining strings and variables:
`print "largest number is: {}".format(max)`

05_syntax_lab/02.py
```diff
@@ -1,6 +1,8 @@
...
-Write a Python program that randomizes 7 numbers
-and prints their sum total.
-If the sum is divisable by 7, also print the word "Boom"
-"""
-
+from random import randint
+sum = 0
+for i in range(7):
+    x = randint(1,100)
+    sum += x
```
ynonp, line 5: Notice how "sum" is also a function name (like "max"), so sum is coloured differently than x

05_syntax_lab/03.py
```diff
@@ -1,7 +1,10 @@
...
-"""
-
+"""
+Take a number from 1 to 10000 and print the sum of its digits
+"""
+from random import randint
+x = str(randint(1,10000))
+print "Number is", x
+sum = 0
+for i in range (len(x)):
```
ynonp, line 8: A text is also a sequence, so you can:
`for character in x: ...`

05_syntax_lab/04.py
```diff
@@ -1,8 +1,11 @@
...
+Read lines and print them
+"""
+lines = []
+while True:
+    line = raw_input()
+    if line == "":
+        break
+    lines.append(line)
+for i in range(0,len(lines)):
+    print lines[len(lines)-1-i]
```
ynonp, line 11: 👍

PR #19, opened 2016-07-05
ToCode Solution: תרגול תחביר

shaiMar:


05_syntax_lab/01.py
```diff
@@ -3,3 +3,14 @@
 the user and prints the largest one
 """
```
ynonp, line 5: Hi Shai and Welcome To The Course

05_syntax_lab/01.py
```diff
@@ -3,3 +3,14 @@
...
 
+largestNumber = int(0)
```
ynonp, line 6: Is 0 really the largest number you expect to read? What if a user types only negative numbers?

05_syntax_lab/01.py
```diff
@@ -3,3 +3,14 @@
...
+
+for n in range(10):
+    print"please enter number(%d)" %(n+1)
+    currentnumber=int(raw_input())
+    if largestNumber < currentnumber:
+        largestNumber = currentnumber
+
+
+
+print('largest number is :'),largestNumber
```
ynonp, line 16: A more pythonic print would probably use str.format:
`print 'largest number is: {}'.format(largestNumber)`

05_syntax_lab/01.py
shaiMar, line 6: so what are you suggesting?

ynonp, line 6: If all inputs are negative then your program would print 0 as the largest value (and not for instance the actual largest value from the input).

One way around this is to differentiate the first value from the rest, and use the first read value as the initial value for `largestNumber`

A second approach is to use the lowest value in the world as the initial value. In python that lowest value is `float('-inf')` (which translates to the negative infinity or מינוס אינסוף)

05_syntax_lab/05.py
```diff
@@ -3,3 +3,15 @@
 until it finds a number that is divisable by: 7, 13 and 15
 """
 
+from random import randint
+
+numberToDiviseto = 7*13*15
```
ynonp, line 8: 👍

05_syntax_lab/05.py
```diff
@@ -3,3 +3,15 @@
...
+
+
+while True:
+    num = randint(1,10000000)
+    if num % numberToDiviseto ==0:
```
ynonp, line 13: In python we prefer lowercase variable names + spaces around equality operators. So prefer `if num % number_to_diviseto == 0`

05_syntax_lab/02.py
```diff
@@ -4,3 +4,15 @@
 If the sum is divisable by 7, also print the word "Boom"
 """
 
+from random import randint
+
+sum = int(0)
```
ynonp, line 9: Prefer `sum = 0` (no need for int(0))

PR #21, opened 2016-07-05

dovron:


06_modules_lab/01.py
```diff
@@ -1,4 +1,17 @@
...
+Python course - Verint
+Week 1
+
+"""
+
+import sys
+
+i = int(sys.argv[1])
+
+while i > 0:
```
ynonp: A for loop is preferred when number of iterations is known in advance (`for i in range(...)`)

06_modules_lab/02_bonus.py
```diff
@@ -1,10 +1,22 @@
...
+
+"""
+from colorama import Fore,init
+import sys
+
+init()
+
+i = sys.argv
+
+if len(i)<>3:
```
ynonp: 1. Prefer using `len(i) != 3` rather than `<>`. The latter is not supported in python3.
2. The function sys.exit provides a better way to exit a script with an error (it tells the OS something bad has happened). Read about it here:
   https://docs.python.org/2/library/sys.html#sys.exit

06_modules_lab/03.py
```diff
@@ -1,11 +1,27 @@
...
+    path = sys.argv[1]
+
+size = 1000000
+if len(sys.argv) == 3:
+    size = int(sys.argv[2])
+
+for root, dir, files in os.walk(path):
+    for name in files:
+        if getsize(join(root, name)) > size:
+            print "Do you want to delete:" , join(root, name),"? [Y/N]"
```
ynonp: string.format can provide a nicer syntax here: `print "Do you want to delete {} ? [y/N]".format(...)`

06_modules_lab/03_bonus.py
```diff
@@ -1,13 +1,36 @@
...
+if args.Size <> "None":
+    size = args.Size
+print size
+
+for root, dir, files in os.walk(path):
+    for name in files:
+        if getsize(join(root, name)) > size:
+            print "Do you want to delete:" , join(root, name),"? [Y/N]"
+            if raw_input() == ("Y"):
+                os.remove(join(root, name))
```
ynonp: 👍

06_modules_lab/02_bonus.py
```diff
@@ -1,10 +1,25 @@
...
+
+if len(i)!=3: 
+    print Fore.RED + 'Error!'
+    print Fore.RESET + "pleae enter 2 number"
+else:
+    try:
+        i = int(sys.argv[1]) + int(sys.argv[2])
+        print "The sum of the numbers is:",i
+    except sys.exit("Error, text is not valid - Please enter a number"):
+        print
```
dovron, line 25: How can I use sys.exit without an unused command after it? (print)

ynonp, line 25: Just write the colon immediately after except. For example:

```
try:
  i = int(sys.argv[1]) + int(sys.argv[2])
  print "..."
except:  
  sys.exit("error invalid text")
```

PR #22, opened 2016-07-05
ToCode Solution: תרגול תחביר

alonshk:


05_syntax_lab/01.py
```diff
@@ -1,5 +1,9 @@
...
-
+
+
+number1 = int(raw_input("Pleas write a number\n"))
+for n in range(9):
+    number2 = int(raw_input("Pleas write a number\n"))
+    if number2 > number1:
+        number1 = number2
+
+print number1
```
szabgab, line 9: Hi and welcome to the exercises!

Please keep the comments at the top of the files so we all know what is the code about.

This seems to work, but could you change the code to eliminate the duplicate call to `raw_input()` ?
Also, you might want to think up some more descriptive names for the variables.

PR #24, opened 2016-07-05
ToCode Solution: תרגול תחביר

moshekenan:


05_syntax_lab/01.py
```diff
@@ -1,5 +1,15 @@
...
-Write a program that reads 10 numbers from
-the user and prints the largest one
-"""
-
+"""
+Syntax excersize #1
+"""
+bigNum = 0
+number = 0
+
```
ynonp, line 6: The program works but may have a surprising result if all input numbers are negative. See if you can figure out what's wrong and fix it.

05_syntax_lab/01.py
```diff
@@ -1,5 +1,15 @@
...
+while number < 10:
+    print "Please enter any number"
```
ynonp, line 8: Bonus: There's an empty line between the print and the input. Can you have both in the same line?

05_syntax_lab/01.py
```diff
@@ -1,5 +1,15 @@
-"""
-Write a program that reads 10 numbers from
-the user and prints the largest one
-"""
-
+"""
+Syntax excersize #1
+"""
+bigNum = 0
```
ynonp, line 4: Variable names in python are usually written lowercase only, so `bignum` is a better name. For long names with multiple words you can use underscores for example: `the_read_huge_number`

05_syntax_lab/02.py
```diff
@@ -1,6 +1,23 @@
-"""
-Write a Python program that randomizes 7 numbers
-and prints their sum total.
-If the sum is divisable by 7, also print the word "Boom"
-"""
-
+"""
+Syntax excersize #2
+"""
+sum = 0
```
ynonp: Notice how github colours "sum" in the same color as print (and differently from "number" and "Boom"). The reason is python has a function named sum, so when you use that as a variable name it can be a bit confusing.

05_syntax_lab/02.py
```diff
@@ -1,6 +1,23 @@
...
+Syntax excersize #2
+"""
+sum = 0
+number = 0
+Boom = ""
+rand = 0
+
+from random import randint
+
+while number < 7:
```
ynonp: A for loop provides a more readable way to perform a block N times (with number of times known in advance). Please modify to use `for` loop

05_syntax_lab/02.py
```diff
@@ -1,6 +1,23 @@
...
+   print rand
+   sum  += rand
+   print sum
+   number += 1
+
+if sum % 7 == 0:
+    Boom = "Boom"
+
+print "The sum of 7 random numbers between 1 and 100 is: {} {}".format(sum, Boom)
+
```
ynonp, line 20: I can see why you used the Boom variable, but I think it turned out not that great. The main problem with a Boom variable is creating a distance between the condition (is sum % 7 == 0) and its effect (the print statement). Reading the code, it's hard to connect when each message is printed and how the condition affects the code that follows.

I think we can have a more readable version by using two variables for the two text strings and using two separate print statements, something like:

```
if sum % 7 == 0:
  print BOOM.format(...)
else:
  print NOBOOM.format(...)
```

Try to define the variables BOOM and NOBOOM without the duplication.

05_syntax_lab/01.py
moshekenan, line 6: Hi Ynon,

Attached fixed program

Moshe Kenan
Software Engineer
Desk: +972 962 4813
Mobile: +972 54 778 4813
Verint Cyber Intelligence
www.verint.comhttp://www.verint.com/

05_syntax_lab/02.py
moshekenan, line 20: Hi Yinon,

Fixed program attached

Moshe Kenan
Software Engineer
Desk: +972 962 4813
Mobile: +972 54 778 4813
Verint Cyber Intelligence
www.verint.comhttp://www.verint.com/

PR #26, opened 2016-07-06

Guyemin:


06_modules_lab/01.py
```diff
@@ -1,4 +1,6 @@
-""" Write a program that takes a count
-from sys.argv import and prints "Hello Python"
-count times
-"""
+import sys
+index = 0
+max = int(sys.argv[1])
+while index < max:
+	print "Hello Python"
+	index = index + 1
```
szabgab, line 6: Nice, but a few comments:
- `max` is a built-in keyword in Python, better use another name
- instead of counting yourself, you could use `range()`

06_modules_lab/02.py
```diff
@@ -1,4 +1,5 @@
-""" Write a program that reads 2 numbers from sys.argv
-and prints their sum.
-"""
-
+import sys
+x = int(sys.argv[1])
+y = int(sys.argv[2])
+sum = x + y
+print "The sum is: %s" % sum
```
szabgab, line 5: 👍

PR #28, opened 2016-07-06
ToCode Solution: תרגול תחביר

ronyblum:


05_syntax_lab/01.py
```diff
@@ -1,5 +1,11 @@
...
+Exercise 1
+
+- Largest number between 10 numbers
+"""
+
+
+num = []
+for i in range(1,11):
+	num.append(raw_input('Please enter the %s number: ' % i))
+print 'The Largest number entered is: ', max(num)
```
szabgab, line 11: Hi Rony and welcome to the exercises!

Nice solution with some advanced tools.
👍

05_syntax_lab/02.py
```diff
@@ -1,6 +1,15 @@
...
+
+from random import randint
+
+sum = 0
+
+for i in range (0,7):
+	i = randint(1, 100)
+	print i
+if i %7 == 0:
+	print 'Boom'
```
szabgab, line 15: You are using the `i` variable for both the iterator and the generated random number. It is confusing.
Oh and you could write `range(7)`
Line 14 is not checking `sum`

PR #29, opened 2016-07-06

EyalAbel:


06_modules_lab/01.py
```diff
@@ -1,4 +1,26 @@
...
+
+
+def Usage():
+    print ('Invalud arguments.\nUsage: 01.py N')
+########################################################################################################################
+#
+#
+def main():
+    if len(sys.argv) != 2:
+        Usage()
```
ynonp, line 15: Prefer lowercase for function names (such as `len`, `sys.argv`, `sys.exit`)

06_modules_lab/03_bonus.py
```diff
@@ -1,13 +1,40 @@
...
+#
+def main():
+    path  = "."
+    parser = argparse.ArgumentParser(description='ArgumentParser use exercise')
+    parser.add_argument("-path", help="path for the working folder")
+    args = parser.parse_args()
+    if args.path:
+        path = args.path
+        if not os.path.isdir(path):
+            print "Error: path must be a valid (existing) directory"
```
ynonp, line 32: Probably better to sys.exit() at this point

06_modules_lab/03_bonus.py
```diff
@@ -1,13 +1,40 @@
...
+    parser = argparse.ArgumentParser(description='ArgumentParser use exercise')
+    parser.add_argument("-path", help="path for the working folder")
+    args = parser.parse_args()
+    if args.path:
+        path = args.path
+        if not os.path.isdir(path):
+            print "Error: path must be a valid (existing) directory" 
+    size = 1024 * 1024 # 1 Megabyte
+    print_file_by_size(size,path)
+    sys.exit()
```
ynonp, line 35: Why do you need to sys.exit here?

06_modules_lab/03_bonus.py
```diff
@@ -1,13 +1,40 @@
...
+def main():
+    path  = "."
+    parser = argparse.ArgumentParser(description='ArgumentParser use exercise')
+    parser.add_argument("-path", help="path for the working folder")
+    args = parser.parse_args()
+    if args.path:
+        path = args.path
+        if not os.path.isdir(path):
+            print "Error: path must be a valid (existing) directory" 
+    size = 1024 * 1024 # 1 Megabyte
```
ynonp, line 33: `size` should also be provided via args

06_modules_lab/03_bonus.py
```diff
@@ -1,13 +1,40 @@
...
-
+#!/usr/bin/env python
+
+__author__ = 'Eabel'
+
+import sys
+import os
+import argparse
+
+def Usage():
```
ynonp, line 9: `argparse` already prints a nice help + usage message when script is used incorrectly

EyalAbel, line 9: I didn't use this function (kept it from previous exercise).

06_modules_lab/03_bonus.py
EyalAbel, line 35: Do real need, just to be explicit that this is the exit point.

PR #30, opened 2016-07-06

razavishay:


06_modules_lab/01.py
```diff
@@ -1,4 +1,4 @@
-""" Write a program that takes a count
-from sys.argv import and prints "Hello Python"
-count times
-"""
+import sys
```
ynonp, line 1: Please leave the top comments in the file (the ones inside """). They make it easier for everyone opening the file later to understand what's inside

06_modules_lab/01.py
```diff
@@ -1,4 +1,4 @@
-""" Write a program that takes a count
-from sys.argv import and prints "Hello Python"
-count times
-"""
+import sys
+userInput = int(sys.argv[1])
+for i in range (1, userInput+1):
```
ynonp, line 3: A simpler `range(userInput)` would suffice here

06_modules_lab/02.py
```diff
@@ -1,4 +1,10 @@
...
+import sys
+try:
+    float(sys.argv[1])
+    float(sys.argv[2])
+except:
+    print "Input must consist of two whole numbers"
+    sys.exit()
+a = float(sys.argv[1])
+b = float(sys.argv[2])
+print "The sum of %f and %f is %f" %(a, b, a+b)
```
ynonp, line 10: Nice! now remove the duplicate `float(...)` statements and we're perfect

06_modules_lab/03.py
```diff
@@ -1,11 +1,17 @@
...
-  a message asking if she wants to delete it, and delete the
-  file if requested
-
-- Take threshold and path as command line arguments
-"""
-
+import os
+import sys
+import argparse
+
```
ynonp, line 4: The code is short and already getting a bit hard to read. Some thoughts here:
1. Use meaningful names in the arguments (path instead of p, size instead of s)
2. Use blank lines between blocks of code: A blank line between the arguments related code and the actual iteration can make a huge difference here

PR #36, opened 2016-07-06

konst1988:


06_modules_lab/01.py
```diff
@@ -2,3 +2,13 @@
...
+
+import sys
+
+if len(sys.argv) > 1:
+    count = sys.argv[1]
+    for i in range(int(count)):
+        print "Hello Python"
+else:
+    program_name = sys.argv[0]
+    print "Usage: %s <count>" % program_name
```
ynonp: A better way to write this condition is to use `sys.exit()`. That also has the advantage to signal to the OS that something went wrong by returning an error code

06_modules_lab/02.py
```diff
@@ -2,3 +2,14 @@
 and prints their sum.
 """
 
+import sys
+
+if len(sys.argv) > 2:
+    (name, strnum1, strnum2) = sys.argv
```
ynonp: Nobody cares about `name` here, so naming it can create confusion. Python has a special variable that holds values nobody cares about - the `_`. A more pythonic way would be:
`(_, strnum1, strnum2) = sys.argv`

PR #37, opened 2016-07-06
ToCode Solution: תרגול תחביר

Oransha:


05_syntax_lab/01.py
```diff
@@ -1,5 +1,18 @@
...
+x5 = int(raw_input ())
+x6 = int(raw_input ())
+x7 = int(raw_input ())
+x8 = int(raw_input ())
+x9 = int(raw_input ())
+x10 = int(raw_input ())
+
+
+print "the largest number is: ",  max (x1,x10)
+
```
szabgab, line 5: Welcome to the exercises!

While the above solution can work, we really don't want to repeat our code 10 times. What if tomorrow, you'll have to solve it for 100 or a 1000?
Please submit a solution that mentions `raw_input()` only once but calls it in a loop 10 times.

05_syntax_lab/01.py
```diff
@@ -1,18 +1,6 @@
...
-x10 = int(raw_input ())
-
-
-print "the largest number is: ",  max (x1,x10)
+print "please insert 10 numbers"
+x = []
+for i in range (10):
+    x.append (int(raw_input()))
 
+print max(x)
```
szabgab, line 6: 👍

05_syntax_lab/02.py
```diff
@@ -1,6 +1,8 @@
...
-"""
-
+# 7 boom
+for i in range (1,8):
+    import random
+    x = random.randint (1,101)
+    sum =+ x
+print "the sum of the random numbers is:", sum 
+if sum % 7 == 0:
+    print "boom"
```
szabgab, line 8: 👍 
The `import random` should always be at the beginning of the program. Outside of any conditions or loops.

05_syntax_lab/05.py
```diff
@@ -1,5 +1,10 @@
...
+import random
+
+while True :
+    x = int(random.randint(1,1000001))  
+    if x % 7 == 0 and x % 13 == 0 and x % 15 == 0:
+        print "this number divisible by 7, 13 and 15:", x
+        break
+    else:
+        continue
+
```
szabgab, line 10: there is no need for the `else` part, but otherwise it is good.

05_syntax_lab/06.py
```diff
@@ -1,9 +1,14 @@
...
+print x, y
+
+for i in range (2,11):
+    if x % i == 0 and y % i == 0:
+        print "the quotient is:", i
+        break
+if i == 10:      
+    print "no solution"
+
+
```
szabgab, line 14: This code searches for the smallest `divider` and not `multiple`.

05_syntax_lab/07.py
```diff
@@ -1,8 +1,14 @@
...
+print "try to guess the number"
+while True:
+    y = int(raw_input())
+    if y == x:
+        print "correct"
+        break
+    if (x-y) > 0:
+        print "the number you guess is too small,try again"
+    else:   
+        print "the number you guess is too big, try again"
```
szabgab, line 14: This is good, but please, use more descriptive variable names. E.g. `hidden` and `guess`

05_syntax_lab/04.py
```diff
@@ -1,8 +0,0 @@
-"""
-Write a program that reads lines from the user
-until an empty line is inserted.
-After the user typed in an empty line,
-print all previously inserted lines in reverse
-order (from last to first)
-"""
-
```
szabgab, line 8: Tthat's an empty file.
In any case you might want to move on to the next set of exercises of the first week.

PR #38, opened 2016-07-06

mashaGC:


06_modules_lab/01.py
```diff
@@ -1,4 +1,8 @@
...
-count times
-"""
+_author_ = 'mariag'
+
+from sys import argv
+
+rep = int(argv[1])
+
+for i in range(rep):
+    print 'Hello Python'
```
szabgab, line 8: Nice, but what happens if the user runs the program without supplying any number?

06_modules_lab/02.py
```diff
@@ -1,4 +1,5 @@
-""" Write a program that reads 2 numbers from sys.argv
-and prints their sum.
-"""
-
+_author_ = 'mariag'
+
+from sys import argv
+
+print int(argv[1]) + int(argv[2])
```
szabgab, line 5: 👍

06_modules_lab/02_bonus.py
```diff
@@ -1,10 +1,16 @@
...
+        int(input)
+        return True
+    except ValueError:
+        return False
+
+if len(argv) != 3 or isInt(argv[1]) == False or isInt(argv[2]) == False:
+    print 'Usage: 02.py <num1> <num2>'
+    exit(1)
+
+print int(argv[1]) + int(argv[2])
```
szabgab, line 16: 👍

PR #41, opened 2016-07-06
ToCode Solution: תרגול תחביר

alonshk:


05_syntax_lab/02.py
```diff
@@ -1,6 +1,9 @@
...
-
+sum = 0
+for i in range(8):
+    from random import randint
+    sum += randint(1, 101)
+print sum
+
+if sum % 7 == 0:
+    print "Boom"
+
```
szabgab, line 9: The import statement should be at the beginning of the file, right after the comment explaining what does the script do.
Please check how many numbers do you generate in this script.

05_syntax_lab/05.py
```diff
@@ -1,5 +1,6 @@
...
-Write a program that randomizes integers in a loop
-until it finds a number that is divisable by: 7, 13 and 15
-"""
-
+
+while True:
+    from random import randint
+    number = randint(1,1000001)
+    if number % 7 == 0 and number % 13 == 0 and number % 15 == 0: break
+
```
szabgab, line 6: Here too, the import needs to be outside.
The program does not print anything.

05_syntax_lab/03.py
```diff
@@ -1,7 +1,14 @@
...
+
+sum = 0
+for i in numberstr:
+    sum += int(i)
+
+print numberstr
+print sum
+
+
+
```
szabgab, line 14: 👍

05_syntax_lab/07.py
```diff
@@ -1,8 +1,20 @@
...
+        print "too BIG;)"
+    elif cube == 2:
+        print "too SMALL;)" #sometime the user will get "too small"
+    elif s > unknown:
+        print "too BIG"
+    elif s < unknown:
+        print "too SMALL"
+    else:
+        print "You are RIGHT"
+        break
```
szabgab, line 20: 👍

PR #42, opened 2016-07-06

zfigov:


06_modules_lab/01.py
```diff
@@ -2,3 +2,9 @@
 from sys.argv import and prints "Hello Python"
 count times
 """
+import sys
+
+if len(sys.argv)==2:
+    times = sys.argv[1]
+    for i in range(0,int(times)):
```
ynonp, line 9: Better to convert to int a line before (when reading from sys.argv[1] to times) and here use: `for i in range(times):`

06_modules_lab/02.py
```diff
@@ -1,4 +1,11 @@
 """ Write a program that reads 2 numbers from sys.argv
 and prints their sum.
 """
+import sys
+
+if len(sys.argv)==3:
+    num1 = sys.argv[1]
+    num2 = sys.argv[2]
+    print "Sum of %s + %s = %d" % (num1,num2,int(num1)+int(num2))
+
```
ynonp, line 10: Why don't you print a nice "usage" message to the user so she knows how to use your script? (in case sys.argv was not the length you expected)

06_modules_lab/03.py
```diff
@@ -8,4 +8,35 @@
...
 - Take threshold and path as command line arguments
 """
+import os
+import readchar
+import sys
 
+correct_input = False
+if len(sys.argv)==3:
+    current_working_directory = sys.argv[1]
+    MinimumSizeInMB = float(sys.argv[2])
```
ynonp, line 18: 1. Use lowercase only variable names (`maximum_size_in_mb` is good, `threshold` is even better)
2. Can also use `(_, x, y) = sys.argv`, which is a bit nicer IMHO

06_modules_lab/03.py
```diff
@@ -8,4 +8,35 @@
...
+
+if correct_input:
+    # size in bytes = 1 MB = 1k * 1k
+    sizeInB = MinimumSizeInMB * 1024 * 1024;
+    for root,dirs,files in os.walk(current_working_directory):
+        for name_f in files:
+           fp = os.path.join(root , name_f)            
+           stat_info = os.stat(fp)
+           if stat_info.st_size> sizeInB:
+               print ("Larger than %f MB %s ") % (MinimumSizeInMB,name_f)
```
ynonp, line 35: no parenthesis around print arguments in python2

06_modules_lab/03.py
```diff
@@ -8,4 +8,35 @@
...
+               print ("Do you want to erase?  Y/N : %s ") % name_f
+               key = readchar.readkey()
+               if key=='Y' or key=='y':
```
ynonp, line 38: `if key.lower() == 'y':` is considered nicer

PR #44, opened 2016-07-07

shaiMar:


06_modules_lab/02.py
```diff
@@ -2,3 +2,13 @@
 and prints their sum.
 """
 
+import sys
+
+if len(sys.argv) < 3:
+    print ('not enough numbers')
```
ynonp, line 8: This is a great place to call `sys.exit(1)`. That command causes the program to end and returns an error code to the operating system.
In addition you won't need the else, as entering the if would end the program. That makes all other program code indented nicely to the left.

PR #45, opened 2016-07-07

moshekenan:


06_modules_lab/01.py
```diff
@@ -1,4 +1,9 @@
-""" Write a program that takes a count
-from sys.argv import and prints "Hello Python"
-count times
-"""
+import sys
+
```
ynonp, line 2: Checking the inverse condition yields nicer code. Consider:

```
if ...:
  print "error"
  sys.exit(1)

# code continues here unindented
```

The advantage is you get some short lines of indented code to handle the error, and most of your program is written nicely aligned to the left

06_modules_lab/02_bonus.py
```diff
@@ -1,10 +1,12 @@
...
-Bonus: Print error messages for invalid inputs.
-
-To print error messages we'll have to use a concept not yet learned in the
-course, and which will only be presented later: Exceptions.
-We'll tap into python's error handling and change its default
-error message to something more meaningful.
-"""
-
+import sys
+
```
ynonp, line 2: Now that you found try/except, you can use them to write cleaner code:
Remove the `if len(sys.argv)...` and wrap everything in a try/except block.
If argument count or type is invalid, you'll end up in the except and can handle it only once (instead of your current code which duplicates the printing of error message)

06_modules_lab/03_bonus.py
```diff
@@ -1,13 +1,35 @@
...
+min_size_to_delete = args.size
+
+print "searching for files in '%s' bigger than %d bytes" % (search_dir, min_size_to_delete)
+
+if os.path.exists(search_dir):
+    for root, dirs, files in os.walk(search_dir):
+        for name in files:
+            #print "found file: %s/%s" % (root, name)
+
+            full_path = "{}\{}".format(root,name)
```
ynonp, line 25: Use `os.path.join(root, name)` to connect file path parts (it is more robust: In unix the path separator is /)

06_modules_lab/03_bonus.py
```diff
@@ -1,13 +1,35 @@
...
+            file_size = os.path.getsize(full_path)
+            if file_size > min_size_to_delete:
+                print "found file %s with size: %d bytes" % (full_path,file_size)
+                print "if you want to delete this file enter 'd'"
+                del_file = raw_input()
+                if del_file == 'd':
+                    os.remove(full_path)
+else:
+    print "The directory %s does not exists" % search_dir
+
```
ynonp, line 35: 👍

PR #46, opened 2016-07-07

shlomitTal:


06_modules_lab/01.py
```diff
@@ -1,4 +1,19 @@
-""" Write a program that takes a count
-from sys.argv import and prints "Hello Python"
-count times
-"""
+#Commited by Shlomit
```
ynonp, line 1: Please leave the top comments in the file (the ones inside """). They make it easier for everyone opening the file later to understand what's inside

06_modules_lab/01.py
```diff
@@ -1,4 +1,19 @@
-""" Write a program that takes a count
-from sys.argv import and prints "Hello Python"
-count times
-"""
+#Commited by Shlomit
+import sys
+if(len(sys.argv)) < 2:
+    print "error - you didn't add any arguments!"
+    sys.exit
```
ynonp, line 5: should be `sys.exit(1)` which returns an error code to the operating system

06_modules_lab/01.py
```diff
@@ -1,4 +1,19 @@
...
-from sys.argv import and prints "Hello Python"
-count times
-"""
+#Commited by Shlomit
+import sys
+if(len(sys.argv)) < 2:
+    print "error - you didn't add any arguments!"
+    sys.exit
+
+loopCount = sys.argv[1]
```
ynonp, line 7: 1. Use lowercase variable names
2. Better to convert to int here: `loop count = int(sys.argv[1])`

06_modules_lab/01.py
```diff
@@ -1,4 +1,19 @@
...
+
+if(loopCount[0] == "-"):
+      print "error - you enter a negetive number!"
+if(not loopCount.isdigit()):
+     print "error - you enter!"
+     sys.exit
+
+i = 0
+while i < int(loopCount):
```
ynonp, line 16: Prefer for loops when the number of iterations is known (`for n in range(...)`)

06_modules_lab/02_bonus.py
```diff
@@ -1,10 +1,14 @@
...
+    sys.exit
+
+arg1 = (sys.argv[1]).lstrip("-+")
+arg2 = (sys.argv[2]).lstrip("-+")
+
+if(not arg1.isdigit() or not arg2.isdigit()):
+     print "error - you dont enter a number!"
+     sys.exit
+
+print "the sum of :",(arg1, arg2), "is:", int(arg1) + int(arg2)
```
ynonp, line 14: Prefer str.format over passing multiple values to print

06_modules_lab/03_bonus.py
```diff
@@ -1,13 +1,29 @@
...
+if not os.path.isdir(path):
+    print "Invalid folder"
+    sys.exit() 
+
+size = args.size
+
+intsize= int(size) * 1024 *1024
+for root, dirs,files in os.walk(path):
+    for name in files:
+        fullFilePathName = os.path.join (root,name)
```
ynonp, line 22: no need for the space. Prefer `os.path.join(root, name)`

PR #48, opened 2016-07-07
ToCode Solution: תרגול קבצים

moshekenan:


08_files_lab/01.py
```diff
@@ -1,6 +1,8 @@
-"""
-Write a program that takes 2 file names
-and appends the second file's contents to
-the end of the first one
-"""
-
+import sys
+
+(src, dst) = sys.argv[1:]
```
ynonp, line 3: 👍

08_files_lab/01.py
```diff
@@ -1,6 +1,8 @@
...
-"""
-
+import sys
+
+(src, dst) = sys.argv[1:]
+
+with open(src, "r") as fin:
+    with open(dst, "a") as fout:
+        for line in fin:
+            fout.write(line)
```
ynonp, line 8: Great for text files, but huge binary files that don't have line breaks will be loaded entirely to memory. Limit buffer size for binary files to make the script more robust.

PR #53, opened 2016-07-07

nirna:


06_modules_lab/01.py
```diff
@@ -1,4 +1,7 @@
...
-from sys.argv import and prints "Hello Python"
-count times
-"""
+""" 
+Print 'Hello Python' K times based on the input argument
+"""
+#Week01_Exercise02_Assignment01
+import sys
+num_of_repeats = int(sys.argv[1])
+for k in range(0,num_of_repeats) : print 'Hello Python'
```
ynonp, line 7: Prefer `range(num_of_repeats)` - the 0 is redundant

06_modules_lab/02_bonus.py
```diff
@@ -1,10 +1,12 @@
...
+"""
+#Week01_Exercise02_Assignment02_Bonus
+import sys
+if len(sys.argv) != 3: print 'Syntax Error. Syntax is:\npython %s <integer_value_1> <integer_value_2>' % sys.argv[0]
+else:
+    try:
+        x,y = int(sys.argv[1]),int(sys.argv[2])
+        print 'Sum of %d and %d is %d' %(x,y,x+y)
+    except ValueError:
+        print 'Syntax Error. Syntax is:\npython %s <integer_value_1> <integer_value_2>' % sys.argv[0]
```
ynonp, line 12: Fantastic! But since you already have try/except no need for the outer if. Just put everything inside the try/except block

06_modules_lab/03.py
```diff
@@ -1,11 +1,16 @@
...
-"""
-
+""" 
+Delete large files (larger from 1MB by default) from specified folder (default is \) 
+"""
+#Week01_Exercise02_Assignment03
+import os
+import sys
+# Read input arguments
+root_dir, size_threshold_bytes = '\\', 1024*1024
```
ynonp, line 8: To keep the script robust, it's better to use the following to get the root path (which works for both unix AND windows):

```
os.path.abspath(os.sep)
```

ynonp:
Lovely. Closing.

PR #55, opened 2016-07-08

eladkah:


06_modules_lab/01.py
```diff
@@ -1,4 +1,12 @@
...
-"""
+'''
+Prints 'Hello Python' message x times when x is received as a parameter to the script
+Prints usage if
+'''
+import sys
+
+if len(sys.argv) == 2 and sys.argv[1].isdigit():
+    arg = int(sys.argv[1])
+    for _ in range(0, arg):
```
ynonp, line 9: `range(arg)` is preferred (no need for the 0)

06_modules_lab/01.py
```diff
@@ -1,4 +1,12 @@
...
+        print 'Hello Python'
+else:
+    print 'Invalid usage. Usage: python 01.py <Number>'
```
ynonp, line 12: What if a user changed the file name? A more robust solution is to use `sys.argv[0]` as the program name

06_modules_lab/02.py
```diff
@@ -1,4 +1,10 @@
...
-"""
-
+'''
+Receives 2 numbers as script parameters and prints their sum.
+Usage: python 02.py <number1> <number2>
+'''
+
+import sys
+
+num1, num2 = (int(sys.argv[1]), int(sys.argv[2]))
```
ynonp, line 8: No need for parenthesis on the right. Python is happy with:  `a, b = c, d`

06_modules_lab/03_bonus.py
```diff
@@ -1,13 +1,34 @@
...
+import os
+import argparse
+
+parser = argparse.ArgumentParser(description='Find big files')
+parser.add_argument('directory', metavar='D', help="Root directory to search")
+parser.add_argument('minSizeInKb', metavar='S', type=int, help="Minimum file size in Kbytes")
+args = parser.parse_args()
+
+directory = args.directory
+minSizeInKb = args.minSizeInKb
```
ynonp, line 17: Prefer lowercase letter names such as `min_size_in_kb`. Even better: `threshold`

06_modules_lab/03_bonus.py
```diff
@@ -1,13 +1,34 @@
...
+
+if not os.path.exists(directory):
+    print '%s does not exists' % directory
+    sys.exit(1)
+
+for root, dirs, files in os.walk(directory):
+    for file in files:
+        sizeInKb = os.stat(root + "\\" + file).st_size / 1024.0
```
ynonp, line 25: Python has `os.path.join` which joins multiple path components, so prefer `os.path.join(root, file)`. That would also work on unix where path separator is `/`

PR #58, opened 2016-07-09
ToCode Solution: תרגול תחביר

shahar476:


05_syntax_lab/01.py
```diff
@@ -1,5 +1,7 @@
...
-the user and prints the largest one
-"""
-
+print "Enter 10 numbers"
+max = int(raw_input())
+for i in range (1,10):
+    number = int(raw_input())
+    if number > max: 
+        high = number
+print "The biggeset number is: %d" % max
```
szabgab, line 7: Quite late, but welcome to the exercises!
- Could you solve this without calling `raw_input()` twice?
- Instead of `%` use the `.format()` for formatting a string.
- Also please rename the variable `max` to something that is not a keyword in Python.

05_syntax_lab/04.py
```diff
@@ -1,8 +1,6 @@
...
-print all previously inserted lines in reverse
-order (from last to first)
-"""
-
+line = raw_input()
+x = ""
+while len(line) > 0:
+    x = line + "\n" + x
+    line = raw_input()
+print x
```
szabgab, line 6: Here too, please eliminate the duplicate call to `raw_input`

PR #59, opened 2016-07-09
ToCode Solution: תרגול קבצים

EyalAbel:


08_files_lab/01.py
```diff
@@ -1,6 +1,42 @@
...
+########################################################################################################################
+#
+#
+def main():
+    parser = argparse.ArgumentParser(description='This program concatinate file_1 into file_2')
+    
+    parser.add_argument("file_1", help="path to file")
+    parser.add_argument("file_2", help="path to file")    
+    args = parser.parse_args()
+    if not is_valid_file_path(args.file_1) or not is_valid_file_path(args.file_2):
```
ynonp: Prefer using try/except over if for detecting errors. This saves you some extra tests and usually results in more correct code. For example - consider what happens if file_2 is not writable.

try/except gives you one place to handle everything that can break, and allows handling different errors differently

08_files_lab/01.py
```diff
@@ -1,6 +1,42 @@
...
+        sys.exit()
+
+    with open(args.file_1, "r") as src:
+        with open(args.file_2, "a") as dst:
+            dst.write(src.read())
```
ynonp: What if file_1 is too large to fit in memory?

08_files_lab/02.py
```diff
@@ -1,7 +1,63 @@
...
+    parser.add_argument("file_1", help="path to file")
+    parser.add_argument("file_2", help="path to file")
+    parser.add_argument("file_3", help="path to file")
+    args = parser.parse_args()
+    if not is_valid_file_path(args.file_1) or not is_valid_file_path(args.file_2):
+        sys.exit()
+    if os.path.isfile(args.file_3):
+        print "Error - file",args.file_3,"alrady exist"
+        sys.exit()
+    if not create_dictionaries(args.file_3):
```
ynonp: Don't turn your exceptions to error codes - use them all the way. Consider just the line `os.makedirs(os.path.dirname(file_path))` inside a try/except block. You'll get the same result but without defining a function and without the if, resulting in much shorter and cleaner code

08_files_lab/02.py
```diff
@@ -1,7 +1,63 @@
...
+        sys.exit()
+    
+
+    with open(args.file_1, "r") as src1:
+        with open(args.file_2, "r") as src2:
+            with open(args.file_3, "w") as dst:        
+                file1_lines = src1.readlines()
+                file2_lines = src2.readlines()
+                file3_lines = filter(None, sum(itertools.izip_longest(file1_lines, file2_lines), ()))
+                dst.writelines(file3_lines)
```
ynonp: With `for` you don't need to read everything into memory

08_files_lab/01.py
EyalAbel: Fixed

08_files_lab/01.py
EyalAbel: Fixed

08_files_lab/02.py
```diff
@@ -9,22 +9,15 @@
...
-    if not os.path.exists(os.path.dirname(file_path)):
-        try:
-            os.makedirs(os.path.dirname(file_path))
-        except OSError as exc: 
-            if exc.errno != errno.EEXIST:           
-                return False
-    return True
+    dir = os.path.dirname(file_path)
+    if not os.path.exists(dir):
+        os.makedirs(os.path.dirname(file_path))
```
ynonp, line 15: 1. You mean create_directories right?
2. P.S. one of my favourite python3 features is:
   https://docs.python.org/3/library/os.html?highlight=makedirs#os.makedirs

08_files_lab/02.py
```diff
@@ -35,29 +28,28 @@ def main():
...
-        
+    try:
+        create_dictionaries(args.file_3)
 
-    
-           
+        file1_lines = load_file(args.file_1)
+        file2_lines = load_file(args.file_2)
+        with open(args.file_3, "w") as dst:
+            itr = itertools.izip_longest(file1_lines, file2_lines)
```
ynonp, line 42: Prefer: `line1, line2 = itertools.izip_longest(file1_lines, file2_lines)`
So you don't need the brackets

PR #60, opened 2016-07-09

EyalAbel:


10_datastructures_lab/01.py
```diff
@@ -1,10 +1,83 @@
...
+        return False
+    return True
+
+def is_valid_ip_address(address):
+    return is_valid_ipv4_address or is_valid_ipv6_address(address)
+
+
+def read_hosts_file(file_path):
+    try:
+        f = open(file_path)
```
ynonp, line 43: Prefer using `with open as f(...)`. That has the advantage of automatically closing the file and clearing resources if anything goes wrong inside the read block

10_datastructures_lab/01.py
```diff
@@ -1,10 +1,83 @@
...
+        host_dict = defaultdict(str)
+        for line in f:
+            items = line.split('=')
+            if len(items) != 2:
+                return False;
```
ynonp, line 48: Don't throw the entire input just because one line failed. Use `continue` to move to the next iteration

10_datastructures_lab/01.py
```diff
@@ -1,10 +1,83 @@
...
+def is_valid_ip_address(address):
+    return is_valid_ipv4_address or is_valid_ipv6_address(address)
+
+
+def read_hosts_file(file_path):
+    try:
+        f = open(file_path)
+        host_dict = defaultdict(str)
+        for line in f:
+            items = line.split('=')
```
ynonp, line 46: Since you're already inside try/except block, let's consider `host, ip = split('=')`.
If it fails the global except is reached and you return false.
If it succeeds you don't need to check `len(items)` And you get to use the much cleaner `host_dict[host] = ip`.
Regarding rstrip - IMHO it's nicer to write it as the first line in the loop, such as: `line = line.rstrip('\r\n')`.

Connecting everything we'd get:

```
for line in f:
  line = line.rstrip()
  host, ip = line.split('=')
  host_dict[host] = ip
```

10_datastructures_lab/01.py
```diff
@@ -1,10 +1,83 @@
...
+        return host_dict
+    except:
+        e = sys.exc_info()[0]
+        print "Error loading hosts file -",e
+        return False
+
+def get_ip(machine,host_dict):
+    ip = host_dict[machine];
+    if not ip:
+        print 'Error -',machine,'is missing from hosts file'
```
ynonp, line 61: Better to write the error message inside `sys.exit()`. It prints the error to stderr AND returns an error code to the operating system.

10_datastructures_lab/01.py
EyalAbel, line 46: Yes but don't you want to check if the IP is valid?

10_datastructures_lab/01.py
EyalAbel, line 61: Nice!

10_datastructures_lab/01.py
ynonp, line 46: That check should also use exceptions: If your `is_valid_ip_address` throws instead of returning false, you can use it as a single line in that loop:

```
for line in f:
    line = line.rstrip()
    host, ip = line.split('=')
    verify_address(ip)
    host_dict[host] = ip
```

And you can decide in the except block if you want to `continue` or `exit`, and main code is kept concise

10_datastructures_lab/02.py
```diff
@@ -1,5 +1,64 @@
...
+########################################################################################################################
+#
+#
+def main():
+    parser = argparse.ArgumentParser(description='This program get words list in a file and prints all of the anagrams')
+    
+    parser.add_argument("file_path", help="path to the word list file")
+    args = parser.parse_args()   
+
+    words = load_words_file(args.file_path)
```
ynonp, line 57: Or consider the alternative: Instead of catching the error in `load_words_file` and printing there, you can drop the try/except from within the function and move it outside to `main`. That way `print_all_anagrams` wouldn't even get called if file was not successfully read.

PR #63, opened 2016-07-09
ToCode Solution: תרגול תחביר

RTsaban:


05_syntax_lab/01.py
```diff
@@ -1,5 +1,11 @@
...
+big = int(raw_input())
+
+for i in range(1,10) :
+    print "Enter Number"
+    num = int(raw_input())
+    big = max(num,big)
+
+print "Max Number is %d",big
+
+
```
szabgab, line 11: Hello,

please eliminate the duplicate call to `raw_input`

RTsaban, line 11: How ? 

i won't able to use max function since big variable is not defined.

szabgab, line 11: Within the loop you can recognize if it is in the first iteration and only assign `big = num`

PR #66, opened 2016-07-10
ToCode Solution: תרגול קבצים

Guyemin:


08_files_lab/01.py
```diff
@@ -1,6 +1,5 @@
...
-Write a program that takes 2 file names
-and appends the second file's contents to
-the end of the first one
-"""
-
+import sys
+with open (sys.argv[1], "r") as fin:
+	with open (sys.argv[2], "w") as fout:
+		for line in fin:
+			fout.write(line)
```
szabgab, line 5: I think 1 and 2 should be the other way around in this exercise.

ynonp, line 5: Hi All and Sorry,

That's actually a translation error on my part. Comments should say the other way around (so this is the correct solution). New fixed versions of the starter is committed.

08_files_lab/02.py
```diff
@@ -1,7 +1,13 @@
...
+b= sys.argv[2]
+c= sys.argv[3]
+with open (c, "w") as ccin:
+	with open (a, "r") as aain:
+		for line in aain:
+			c = ccin.write(line)
+		c = ccin.write("\n")
+		with open (b, "r") as bbin:
+			for line in bbin:
+				c = ccin.write(line)
```
szabgab, line 13: I tried with two input files

```
a1
a2
```

and

```
b1
b2
```

and was expecting

```
a1
b1
a2
b2
```

but got

```
a1
a2

b1
b2
```

besides, I think the name ccin is a bit misleading for an output file.

PR #68, opened 2016-07-10

EyalAbel:


14_functions_lab/01.py
```diff
@@ -1,7 +1,36 @@
...
+
+def mysum(*nums):
+    """
+    Args:
+        nums (int list): numbers to accumulate (ignore non integers arguments)
+    Returns:
+        int: Return an sum of the arguments
+    """
+    sum = 0
+    for num in nums:
```
ynonp, line 19: Now with list comprehensions:

```
def mysum(*nums):
    return reduce(operator.add, [n for n in nums if isinstance(n, numbers.Number)])
```

14_functions_lab/04.py
```diff
@@ -1,6 +1,23 @@
...
+
+def longer_than(size,*words):
+    """
+    Args:
+        size (int): size of the min word length
+        words (string list): list of words
+    Returns:
+        int: subset of 'words' list that are longer then 'size'
+    """
+    if not all(isinstance(item, basestring) for item in words) or not isinstance(size, numbers.Number):
```
ynonp, line 18: Do you really need this validation? The list comprehension will explode for inputs that don't have `len()`, and without the validation you get the extra functionality that the same code can now check other things with `len`

14_functions_lab/05.py
```diff
@@ -1,10 +1,20 @@
...
+from collections import defaultdict
+
+
+def groupby(key_func,*values):
+    """
+    Returns:
+    
+        dictionary:  grouped by value of key_func(v) (for v in values)
+    """
+    d = defaultdict(list)
```
ynonp, line 17: ⭐

PR #69, opened 2016-07-10
ToCode Solution: תרגול תחביר

idisis:


05_syntax_lab/01.py
```diff
@@ -1,5 +1,6 @@
-"""
-Write a program that reads 10 numbers from
-the user and prints the largest one
-"""
-
+print 'Enter 10 integers, 1 in each line:'
```
ynonp, line 1: Hi Nir and welcome to the course,

Please leave the top comments in the file (the ones inside """). They make it easier for everyone opening the file later to understand what's inside

05_syntax_lab/02.py
```diff
@@ -1,6 +1,7 @@
-"""
-Write a Python program that randomizes 7 numbers
-and prints their sum total.
-If the sum is divisable by 7, also print the word "Boom"
-"""
-
+from random import randint
+
+sum = int(0)
```
ynonp, line 3: Notice how github colours "sum" in the same color as print (and differently from "number" and "Boom"). The reason is python has a function named sum, so when you use that as a variable name it can be a bit confusing.

ynonp, line 3: Also no need for `int(0)`. Just `x = 0` is enough

05_syntax_lab/04.py
```diff
@@ -1,8 +1,10 @@
...
-"""
-
+lines = list()
+i = int(0)
+while True:
+  i += 1
+  line = raw_input("line {}: ".format(i))
+  if not line: break
+  lines.append(line)
+lines.reverse()
```
ynonp, line 8: python's lists have `lines.insert(0, value)` which adds to the beginning

05_syntax_lab/05.py
```diff
@@ -1,5 +1,6 @@
-"""
-Write a program that randomizes integers in a loop
-until it finds a number that is divisable by: 7, 13 and 15
-"""
-
+from random import randint
+
+while True:
+    number = randint(1, 1000000)
+    if number % (7*13*15) == 0: break
```
ynonp, line 5: 👍

PR #70, opened 2016-07-10

HananRosenthal:


10_datastructures_lab/01.py
```diff
@@ -1,10 +1,21 @@
...
+
+table = {'apple': 'red', 'lettuce': 'green', 'lemon': 'yellow', 'orange': 'orange'}
+
+if kuku.string1 in table:
+    if table[kuku.string1] == kuku.string2:
+        print('Welcome')
+        print(u'\u05d1\u05e8\u05d5\u05db\u05d9\u05dd \u05d4\u05d1\u05d0\u05d9\u05dd')
+    else:
+        print('Failed')
+        print(u'\u05d0\u05d9\u05df \u05db\u05e0\u05d9\u05e1\u05d4')
```
szabgab, line 21: 👍

10_datastructures_lab/03.py
```diff
@@ -1,5 +1,26 @@
...
+while True:
+    strr = input('Please enter a device name, enter to exit')
+    if strr == "":
+        break
+    elif strr in hosts:
+        print(hosts[strr])
+    else:
+        print('Device {} not found'.format(strr))
+
+print('Thank you and goodbye')
```
szabgab, line 26: 👍 your variable names are interesting :)

10_datastructures_lab/02.py
```diff
@@ -1,5 +1,15 @@
...
+    grades.append(randint(1,100))
+
+print('The grades are:{}'.format(grades))
+
+avg = float(sum(grades)/len((grades)))
+print('The average grade is {}'.format(avg))
+
+for i in grades:
+    if i > avg:
+        print(grades.index(i))
```
szabgab, line 15: 👍

10_datastructures_lab/04.py
```diff
@@ -1,7 +1,43 @@
...
+    length = len(keys)
+    if length > 0:
+        print('The following words are anagrams\n\r')
+        print('{}'.format(iam[0]))
+        for i in range(0,length):
+            print('{}'.format(keys[i]))
+
+    for i in keys:
+        del word_dic[i]
+    keys.clear()
```
szabgab, line 43: This looks complicated. I'd probably sort the characters in each word and then compare the sorted versions. In anagrams they should be equal.

HananRosenthal, line 43: how would u do that?

szabgab, line 43: `''.join(sorted("acb"))`  returns the string with the sorted letters. This can be used as the key in a dictionary and the value can be a list of the original strings.

PR #72, opened 2016-07-10

Oransha:


06_modules_lab/01.py
```diff
@@ -1,4 +1,13 @@
...
+    print "Error, please enter only one number"
+    sys.exit()
+
+num = int(sys.argv[1])
+
+for i in range (0,num):
+    print "Hello python"
+
+    
+
```
szabgab, line 13: 👍

06_modules_lab/02.py
```diff
@@ -1,4 +1,10 @@
...
+import sys
+
+if len(sys.argv) != 3:
+    print "Error, please enter only 2 numbers"
+    sys.exit()
+
+num1 = int(sys.argv[1])
+num2 = int(sys.argv[2])
+
+print "the sum is:", (num1+num2)
```
szabgab, line 10: 👍

06_modules_lab/03.py
```diff
@@ -1,11 +1,40 @@
...
+            while True:
+                if user_call == "y":
+                    os.remove(file)
+                elif user_call == "n":
+                    break
+                else:
+                    print "error, your input is wrong, try again"
+                    user_call = raw_input()
+
+
```
szabgab, line 40: It is better to prompt like this `raw_input('text')` instead of the separate print statement.
If you used argparse ( https://docs.python.org/2/library/argparse.html ) to get the parameters then you can make both of them optional with default values and you don't need to have the additional questions as you have now.

PR #73, opened 2016-07-10

EyalAbel:


18_regexp_lab/01.py
```diff
@@ -1,12 +1,35 @@
...
+#
+def main():
+    parser = argparse.ArgumentParser(description='This program receive config file path and varible name and prints the varible value')
+    
+    parser.add_argument("config_path", help="path to config file")
+    parser.add_argument("varible_name", help="varible name")
+    args = parser.parse_args()
+
+    try:
+        regex_str = '\s*' + args.varible_name + '\s*=\s*(.*)'
```
ynonp, line 22: Better not to use external input as regular expression. Many characters have special meaning in regexp context (such as: $, ^, \, *, ., etc.)

You can get better results here by reading the file into a dictionary (split every line and use the values as key/value pair), and then search the dictionary for the required value.

Also beware of '#' lines which are comments

18_regexp_lab/03.py
```diff
@@ -1,15 +1,36 @@
...
+                columns[0] = columns[1]
+                columns[1] = tmp
+                columns[2] = columns[2].rstrip('\n')
+                print ','.join(columns)
+    except:
+        e = sys.exc_info()[0]
+        print "Error loading file -",e
+                         
+if __name__ == '__main__':
+    main()
```
ynonp, line 36: Bonus 1: Search for `re.sub` and see how it can simplify your code
Bonus 2: Try to integrate module:
https://docs.python.org/2/library/csv.html

18_regexp_lab/01.py
EyalAbel, line 22: Sure I thinks so too, but it was regex practice so...

ynonp, line 22: I was referring to re.split (to handle whitespace):
https://docs.python.org/2/library/re.html#re.split

Also you need a regex to ignore comments

PR #74, opened 2016-07-10

EyalAbel:


20_classes_lab/03.py
```diff
@@ -1,13 +1,52 @@
...
+
+    def add_dependency(self,*modules):
+        for module in modules:
+            self._dependencies.add(module._name)
+            self._dependencies.update(module._dependencies)
+            
+
+    def build(self):
+        print ', '.join(self._dependencies)
+
```
ynonp, line 25: I don't think this works as requested.

Dependencies should be printed in an order such that no module is printed before all of its dependencies are printed.

PR #75, opened 2016-07-10
ToCode Solution: תרגול קבצים

konst1988:


08_files_lab/01.py
```diff
@@ -4,3 +4,12 @@
...
 
+import sys
+
+(src, dst) = sys.argv[1:]
+
+with open(src, "r") as fin:
+    with open(dst, "a") as fout:
+        for line in fin:
+            fout.write(line.title())
+
```
ynonp: Cool! But what if src is a huge binary file with no line breaks? Python then tries to read the whole file into memory and may fail. Can you update the program so it won't depend on line breaks to work properly?

ynonp:
Great Work. Closing.

PR #76, opened 2016-07-10
ToCode Solution: תרגול תחביר

idanpp:


05_syntax_lab/01.py
```diff
@@ -1,5 +1,10 @@
...
+
+print "Please choose Number"
+x = int(raw_input())
+
+for number in range(1,10) :
+    print "choose agian"
+    num = int(raw_input())
+    x = max(num,x)
+
+print "Highes number is" ,x
```
szabgab, line 10: Hi and a late welcome to the exercises!

You could eliminate the duplicate call to ``raw_input`.

05_syntax_lab/02.py
```diff
@@ -1,6 +1,10 @@
...
+from random import randint
+    
+sum = 0
+for x in range(1,7) :
+    num = randint(1,100)
+    sum += num
+
+print sum
+if (sum % 7) == 0 :
+    print "BOOM"
```
szabgab, line 10: range in python includes the low number but not the high number so range(1, 7) will cover 1,2,3,4,5,6.

05_syntax_lab/03.py
```diff
@@ -1,7 +1,11 @@
...
+
+sum = 0
+
+num = randint(1,10000)
+print num 
+while num > 0 :
+    diggit = num % 10 
+    sum = sum + diggit
+    num = num / 10 
+print sum
```
szabgab, line 11: 👍

05_syntax_lab/04.py
```diff
@@ -1,8 +1,20 @@
...
+
+
+
+
+
+
+
+
+
+
```
szabgab, line 20: <> is supported in Python 2 but not in Python 3. Better to use != instead.

05_syntax_lab/05.py
```diff
@@ -1,5 +1,9 @@
...
-
+from random import randint
+
+while True:
+    num = randint(1,1000000)
+    if num % 7 == 0:
+        if num % 13 == 0:
+            if num % 15 == 0:
+                 print num
+                 break;
```
szabgab, line 9: This could be better written in a single if statement connection the conditions with 'and'.

PR #77, opened 2016-07-10
ToCode Solution: תרגול קבצים

eladagam:


08_files_lab/01.py
```diff
@@ -1,6 +1,25 @@
...
+source = os.path.join(workingDir, args['source'])
+destination = os.path.join(workingDir, args['destination'])
+
+
+with open(destination,"a") as fout:
+    with open(source,"r") as fin:
+        for line in fin:
+            fout.write(line)
+
+
```
szabgab, line 25: Appending the filename to the current directory is not necessary and actually might be even problematic if the user gives absoulte path from the root of the filesystem. Like this

`python 01.py /tmp/a  /tmp/b`

szabgab, line 25: I liked the use of argparse!

08_files_lab/02.py
```diff
@@ -1,7 +1,36 @@
...
+#Merge two input files to the output file
+with open(destination,"w") as fout:
+    with open(source1,"r") as fin1:
+        with open(source2,"r") as fin2:
+            for i in itertools.izip_longest(fin1, fin2):
+                if i[0]!=None:
+                    fout.write(i[0])
+                if i[1]!=None:
+                    fout.write(i[1])
+
```
szabgab, line 36: 👍 for using argparse
Here too better leave the path as the user gave it.
It is better to leave spaces around operators such as `!=`. It makes the code more readable.

PR #78, opened 2016-07-10

moshekenan:


10_datastructures_lab/01.py
```diff
@@ -1,10 +1,23 @@
...
+user_name = raw_input("Please enter username: ")
+password = raw_input("Please enter password: ")
+
+#return_str = "access denied"
+return_str = u"אין כניסה"
+
+#check validity
+for valid in valid_accounts:
+    if user_name == valid and password == valid_accounts[valid]:
+        #return_str = "Welcome"
```
ynonp: No need to iterate over a dictionary. That's the nice thing about it. Consider:

```
if user_name in valid_accounts and valid_accounts[user_name] == password:
    # Welcome!
```

10_datastructures_lab/02.py
```diff
@@ -1,5 +1,22 @@
...
+from random import randint
+
+sum_of_grades = 0
+
+grades = range(20)
+
+for n in grades:
+   grades[n] = randint(40, 100)
+   sum_of_grades += grades[n]
+
```
ynonp: Better to use python's sum: `sum_of_grades = sum(grades)`

10_datastructures_lab/01.py
```diff
@@ -1,10 +1,31 @@
...
+excersize #1
+Print IP address of hosts
+"""
+
+import sys
+
+hosts = {}
+
+with open("G:\People\Moshe\hosts.txt", "r") as fin:
+    lines = [line.rstrip('\n') for line in fin]
```
ynonp, line 11: That's just overusing list comprehension :)
The down side is you read the file twice (taking twice as long), so better to rstrip() inside the second loop

10_datastructures_lab/01.py
```diff
@@ -1,10 +1,31 @@
...
+    for line in lines:
+        fields = line.split("=")
```
ynonp, line 13: Consider:

```
host, ip = line.rstrip().split('=')
```

10_datastructures_lab/02.py
```diff
@@ -1,5 +1,11 @@
-"""
-Write a program that prints only unique values
-from sys.argv (without program name)
-"""
-
+"""
+excersize #1
+Print list of all possible 3 lowercase English letters sequence
+"""
+
```
ynonp, line 5: I think you uploaded the wrong file here

PR #80, opened 2016-07-11
ToCode Solution: תרגול קבצים

Oransha:


08_files_lab/01.py
```diff
@@ -1,6 +1,12 @@
...
+
+(_, file1, file2) = sys.argv
+
+with open ("./" + file1, 'r') as fin:
+    with open ("./" + file2, 'w') as fout:  
+        for line in fin:
+            fout.write(line)
+            print line
+
+#should I delete file1 contents or just copy the lines to file2?
```
szabgab, line 12: Just copy is ok, but appending the provided filename to './' is not a good idea. What if the user gives you /tmp/a and /tmp/b ? Will it still work?

08_files_lab/02.py
```diff
@@ -1,7 +1,27 @@
...
+"""" my ootput is : 
+one
+red
+two
+blue
+threegreen
+white
+orange
+
+how can i split the str "threegreen" ? """"
```
szabgab: What does the `write` function do here?
You need to add a newline after printing "three", but what is the general case of this problem?

08_files_lab/01.py
Oransha, line 12: so it's better to ask the user to insert full path ?

szabgab, line 12: No need for that. Try the code with relative and with full path and see how does it work.

Oransha, line 12: Maybe I'm misunderstanding but when I'm writing full path (without the file name because the user need to select it), I'm limiting the user to a specific path, 
lets say:

**with open( "c:\" + file1 .......)**
so the limitation here is that the program is searching files only in disk C

szabgab, line 12: The best is to write `open(file1)`    so the user can pass   path/to/file.txt   or c:\path\to\file.txt or ....\other\path\file.txt

PR #81, opened 2016-07-11
ToCode Solution: תרגול קבצים

tomag:


08_files_lab/01.py
```diff
@@ -1,6 +1,18 @@
...
-
+"""
+The program receives two files as input
+and writes the content of the first file to the second one
+"""
+import sys
+
+if len(sys.argv) != 3:
+    print "Usage: 01.py <firstFile> <secondFile>"
+    sys.exit()
```
ynonp, line 9: Better to write the error message inside `sys.exit()`. It prints the error to stderr AND returns an error code to the operating system.

08_files_lab/02.py
```diff
@@ -1,7 +1,17 @@
...
+    sys.exit()
+
+(first,second,output) = sys.argv[1:]
+
+with open(first, "r") as firstFile:
+    with open(second, "r") as secondFile:
+        with open(output, "w") as outputFile:
+            for line in itertools.izip_longest(firstFile, secondFile, fillvalue=""):
+                outputFile.write(line[0])
+                outputFile.write(line[1])
```
ynonp, line 17: ⭐

PR #83, opened 2016-07-11

idisis:


06_modules_lab/01.py
```diff
@@ -1,4 +1,5 @@
-""" Write a program that takes a count
-from sys.argv import and prints "Hello Python"
-count times
-"""
+from sys import argv
+
+(script_name, lines) = argv
```
ynonp, line 3: Since nobody uses `script_name` it's better to name it `_` (this is python's convention for a variable nobody cares about)

06_modules_lab/02.py
```diff
@@ -1,4 +1,12 @@
-""" Write a program that reads 2 numbers from sys.argv
-and prints their sum.
-"""
-
+from sys import argv
+
+if len(argv) != 3:
+    print 'Usage:\n%s [a] [b]' %argv[0]
```
ynonp, line 4: Better to write the error message inside `sys.exit()`. It prints the error to stderr AND returns an error code to the operating system.

06_modules_lab/02.py
```diff
@@ -1,4 +1,12 @@
...
+
+if len(argv) != 3:
+    print 'Usage:\n%s [a] [b]' %argv[0]
+    exit()
+
+try:
+  a = int(argv[1])
+  b = int(argv[2])
+  print 'The sum of %d and %d is %d.' %(a, b, a+b)
+except ValueError:
```
ynonp, line 11: Better to use exception all the way. Specifically, you can remove the first `if` and move all invalid input handling to the except block.
It doesn't matter if a user typed in the wrong number of inputs or their wrong type.

06_modules_lab/01.py
idisis, line 3: Thanks! I was actually wondering about the _ usage in python scripts... That's very helpful :+1:

PR #85, opened 2016-07-11
ToCode Solution: תרגול קבצים

omerblecher:


08_files_lab/01.py
```diff
@@ -1,6 +1,9 @@
...
-
+#Append file 1 into the end of file 2
+import sys
+
+(src, dst) = sys.argv[1:]
+
+with open(src, "r") as fin: #Open file for reading
+    with open(dst, "a+") as fout: #Open file for appending. If the file doesn't exist then create it
+        for line in fin:
+            fout.write(line) #Copy each line
```
szabgab, line 9: 🍪 for taking from argv only what you need.
Better not to add that + sign after the "a" in the opening as this means the file is open for reading as well. Which we don't need.

08_files_lab/02.py
```diff
@@ -1,7 +1,13 @@
...
+
+(firstSource, secondSource, dst) = sys.argv[1:]
+
+with open(firstSource, "r") as finFirst: # Open first file for reading
+    with open(secondSource, "r") as finSecond: # Open second file for reading
+        with open(dst, "w+") as fout: # Open file only for writing. Truncate it if needed
+            for step in itertools.izip_longest(finFirst, finSecond): #Get tuple from two files
+                for item in step:
+                    if item is not None: #Write it to the file only if it is not none
+                        fout.write(item)
```
szabgab, line 13: 👍 nice.

PR #86, opened 2016-07-11
ToCode Solution: תרגול תחביר

nirbar1985:


05_syntax_lab/01.py
```diff
@@ -1,5 +1,19 @@
...
+
+write program that gets from user 10 numbers and prints the biggest
+
+
+"""
+
+
+max =0
+
+for i in range (1,11):
```
ynonp, line 11: Since you don't care about the value of i inside the loop, there's a nicer syntax for this:

```
for _ in range(10):
```

05_syntax_lab/01.py
```diff
@@ -1,5 +1,19 @@
...
-
+"""
+
+write program that gets from user 10 numbers and prints the biggest
+
+
+"""
+
+
+max =0
```
ynonp, line 9: If you start with `float("-inf")` all negative inputs will also work

05_syntax_lab/01.py
```diff
@@ -1,5 +1,19 @@
...
+
+for i in range (1,11):
+
+	num = int(raw_input())
```
ynonp, line 13: raw_input() can take an argument and then uses it as a prompt value, so better to use: `num = int(raw_input("Type your number: "))`

05_syntax_lab/01.py
```diff
@@ -1,5 +1,19 @@
...
+	if (num>max):
+		max=num
```
ynonp, line 15: Notice how github colours "max" in the same color as print (and differently from "num"). The reason is python has a function named max, so when you use that as a variable name it can be a bit confusing.

05_syntax_lab/03.py
```diff
@@ -1,7 +1,17 @@
...
+
+
+"""
+
+
+from random import randint
+num = randint(1, 10000)
+
+list_of_ints = [int(i) for i in str(num)]
+sum = 0
```
ynonp, line 13: sum is already a python function, so you can simply:
`print sum(list_of_ints)`

05_syntax_lab/02.py
```diff
@@ -1,6 +1,19 @@
...
+from random import randint
+
+sum=0
+
+for i in range (1,11):
+
+	sum +=randint(1, 100)
+print sum
+
+if(sum%7 == 0):
```
ynonp, line 18: No need for parenthesis in python. This works: `if sum % 7 == 0:`

05_syntax_lab/06.py
```diff
@@ -1,9 +1,22 @@
...
+
+num1 = randint(1, 10)
+num2 = randint(1, 10)
+
+i=0
+
+while(i != num1*num2):
+
+	i += num1
+	if(i%num2==0):
```
ynonp, line 17: Better to write: `if i % num2 == 0: break`

PR #89, opened 2016-07-11

Guyemin:


10_datastructures_lab/01.py
```diff
@@ -1,10 +1,5 @@
...
-    apple => red
-    lettuce => green
-    lemon => yellow
-    orange => orange
-"""
+import sys
+ip= {'work':'10.0.0.2', 'router':'10.0.0.1', 'mycard': '10.0.0.5', 'router':'10.0.60.6'}
+for arg in sys.argv:
+	if arg in ip:
+		print ip[arg]
```
szabgab, line 5: Hmm, this code does not seem to be related to the exercise. What happened?
Update: I see the Hebrew description and the English were not the same here either.
Try to read the hostname => IP address mappings from a file that looks like this:

```
work=10.0.0.2
router=10.0.0.1
mycar=10.0.0.5
home=194.90.2.1
```

PR #90, opened 2016-07-11

moshekenan:


12_listcomprehension_lab/01.py
```diff
@@ -1,8 +1,8 @@
...
-
-
+"""
+excersize #1
+Print list of all English lowercase letters
+"""
+
+x = range(ord('a'),ord('z')+1)
+for i in x: print chr(i)
+
```
ynonp, line 9: Please use list comprehension to generate the list:
https://www.tocode.co.il/bundles/python-verint/lessons/list-comprehension

12_listcomprehension_lab/01.py
```diff
@@ -1,8 +1,9 @@
...
-
-
-
+"""
+excersize #1
+Print list of all English lowercase letters
+"""
+
+#x = range(ord('a'),ord('z')+1)
+lowercase_letters = [x for x in range(256) if x >= ord('a') and x <= ord('z')]
```
ynonp, line 7: Prefer to use `ord('a')` and `ord('z')` as your limits in range, so you don't need the if

PR #92, opened 2016-07-11
ToCode Solution: תרגול קבצים

dovron:


08_files_lab/01.py
```diff
@@ -1,6 +1,17 @@
...
+Python course - Verint
+Week 2
+
+"""
+
+import sys
+
+with open(sys.argv[1], "r") as fin:
+    with open(sys.argv[2], "a") as fout:
+        fout.write("\n")
```
ynonp, line 15: Why do you need to start with a newline?

dovron, line 15: In order to have seperation between the files,
Without using the newline the output will be like:

image

08_files_lab/02.py
```diff
@@ -1,7 +1,18 @@
...
+"""
+
+import sys
+import binascii
+from itertools import izip_longest
+
+with open(sys.argv[1], "r") as fin1:
+    with open(sys.argv[2], "r") as fin2:
+        with open(sys.argv[3], "w") as fout:
+            print >> fout, list(izip_longest(fin1,fin2,fillvalue=None))
```
dovron: Hi,
Could you give more example about izip_longest function?

How can i fix my output?
image

dovron: ?

08_files_lab/01.py
dovron, line 15: ?

ynonp, line 15: Only if the file does not end with a '\n' (which is NOT commonly the case). Many editors will automatically add a '\n' to the end of the last line in the file.
If you insist on handling files that don't have a newline at their end it's better to check for the (non)existence of the newline to prevent double newline problem

08_files_lab/02.py
ynonp: Use a for loop:

```
for line1, line2 in izip_longest(fin1, fin2, fillvalue=""):
  print >>fout, line1 + line2
```

PR #93, opened 2016-07-12
ToCode Solution: תרגול קבצים

Guyemin:


08_files_lab/01.py
```diff
@@ -1,6 +1,5 @@
...
-Write a program that takes 2 file names
-and appends the second file's contents to
-the end of the first one
-"""
-
+import sys
+with open (sys.argv[1], "r") as fin:
+	with open (sys.argv[2], "w") as fout:
+		for line in fin:
+			fout.write(line)
```
szabgab: 👍

08_files_lab/02.py
```diff
@@ -1,7 +1,13 @@
...
+b= sys.argv[2]
+c= sys.argv[3]
+with open (c, "w") as ccin:
+	with open (a, "r") as aain:
+		for line in aain:
+			c = ccin.write(line)
+		c = ccin.write("\n")
+		with open (b, "r") as bbin:
+			for line in bbin:
+				c = ccin.write(line)
```
szabgab, line 13: As I can see this would write the files one after the other and not mixed. Given two files

```
a1
a2
```

and

```
b1
b2
```

We would like to get a new file with

```
a1
b1
a2
b2
```

PR #97, opened 2016-07-12
ToCode Solution: תרגול קבצים

shaiMar:


08_files_lab/02.py
```diff
@@ -4,4 +4,34 @@
...
+import sys
+
+file_name1 = sys.argv[1]
+file_name2 = sys.argv[2]
+output_filename = sys.argv[3]
+
+with  open(file_name1,'r') as f1:
+    with open(file_name2, 'r') as f2:
+        with open(output_filename, 'w') as fout:
+            l1 = f1.readline()
```
ynonp, line 16: That's exactly what izip_longest does :)

Try to run the following (and then modify your code to use that function):

```
import itertools

l1 = [10, 20, 30]
l2 = ['foo', 'bar']

for num, word in itertools.izip_longest(l1,l2):
    print "Num = {}, Word = {}".format(num,word)
```

PR #100, opened 2016-07-12
ToCode Solution: תרגול תחביר

gildahan:


05_syntax_lab/01.py
```diff
@@ -1,5 +1,13 @@
-"""
-Write a program that reads 10 numbers from
-the user and prints the largest one
-"""
-
+"""
+Print thr bigest number between 10 inputs
+"""
+newX=0
```
ynonp, line 4: Hi Gil And Welcome,

This works great when some inputs are positive, but if all inputs are negative you'll get the wrong result.

Can you find the absolute lowest value in python?

05_syntax_lab/02.py
```diff
@@ -1,6 +1,17 @@
...
-Write a Python program that randomizes 7 numbers
-and prints their sum total.
-If the sum is divisable by 7, also print the word "Boom"
-"""
-
+from random import randint
+"""
+Print the sum of a 7 random number between 1-100
+"""
+newX=0
```
ynonp, line 5: Python variable names should be written lowercase only

05_syntax_lab/03.py
```diff
@@ -1,7 +1,13 @@
...
-The result should be: 14
-"""
-
+from random import randint
+"""
+Print the sum of a the digits of a random number between 1-10000
+"""
+
+
+input=str(randint(1,10000))
```
ynonp, line 7: Notice how github colours "input" in the same color as print (and differently from "onechar"). The reason is python has a function named input, so when you use that as a variable name it can be a bit confusing.

05_syntax_lab/03.py
```diff
@@ -1,7 +1,13 @@
...
+print "Selected random number as string "  , input
+sum=0
+for oneChar in input:
+    sum += int(oneChar)
+
+print "The sum of the digits is : " ,sum
```
ynonp, line 13: string.format is generally preferred for printing: `print "The sum is {}".format(sum)`

05_syntax_lab/04.py
```diff
@@ -1,8 +1,18 @@
...
+line = raw_input()
+lines.insert(0, line)
+
+while line != '': 
+        line = raw_input()
+        lines.insert(0, line)
+        
+
+for l in lines:
+        print  l
```
ynonp, line 18: Nice. Can you also solve this without using arrays?

05_syntax_lab/05.py
```diff
@@ -1,5 +1,14 @@
...
+Finding a number deviding 15 , 13 ,7
+"""
+
+num = randint(1,1000000)
+while (num % 7 != 0) or (num % 13 != 0) or (num % 15 != 0): 
+    print "The current selected numer is : " , num
+    num = randint(1,1000000)
+    
+print "The winner num is :", num
+
```
ynonp, line 14: 👍

PR #101, opened 2016-07-12
ToCode Solution: תרגול קבצים

dorsegal:


08_files_lab/02.py
```diff
@@ -1,7 +1,16 @@
...
+parser.add_argument("srcfile2", type=str ,help="the second source file")
+parser.add_argument("dstfile", type=str, help="destination file containing marge content")
+
+args = parser.parse_args()
+
+with open(args.srcfile1, "r") as srcfile1:
+    with open(args.srcfile2, "r") as srcfile2:
+        with open(args.dstfile, "w") as dstfile:           
+            for line, line2 in izip_longest(srcfile1, srcfile2, fillvalue=""):
+                dstfile.write(line + line2)
```
DorSegal: how can i overcome 
image

besides writing extra line in the end of the file on file 1?

ynonp, line 16: If each file ends with a line break that error shouldn't happen. If you do wish to handle files that have no line break at the end, you can check line.endswith('\n') and print an extra '\n' for lines that don't have EOL char

PR #105, opened 2016-07-12

Oransha:


10_datastructures_lab/01.py
```diff
@@ -1,10 +1,19 @@
...
+from sys import argv
+for hostname in argv:
+    if hosts.has_key(hostname):
+        print hosts[hostname]
+    elif hostname == program_name:
+        continue
+    else:
+        print "the hostname you entered is invalid"
+    
+
```
szabgab, line 19: A couple of issues:
- There is no need for ./  , it is clearer to write `open('hosts.txt', 'r')`
- Lines 8-9 are better written like this:

```
        name, ip = line.split("=")
       hosts[name] = ip
```
- Instead of `hosts.has_key(hostname)` it is nicer to write `hostname in hosts`
- You could iterate over `for hostname in argv[1:]` and then there is no need for the 14-15 lines which are confusing anyway.

Oransha, line 19: Thanks for the detailed feedback!

10_datastructures_lab/02.py
```diff
@@ -1,5 +1,14 @@
...
+with open(file1, 'r') as fin:
+    for word in fin:
+        word = word.strip('\n')
+        key = sorted(word)
+        key_str = ''.join(key)
+        word_dict.setdefault(key_str, [])
+        if key_str in word_dict:
+            word_dict[key_str].append(word)
+    for key_str in word_dict:
+        print word_dict[key_str]
```
szabgab, line 14: 👍 
You could eliminate the _key_ variable and merger the lines 8-9
Instead of setdefault and lines 9-12 I'd probably write:

```
if key_str in word_dict:
    word_dict[key_str].append(word)
else:
    word_dict[key_str]= [word]
```

Oransha, line 14: OK, thanks.
I just want to ask you, I wasn't able to find solution to this problem without "join".
There is another way ?

szabgab, line 14: You could also use tuples of all the sorted characters, but I like the one with join better.

PR #107, opened 2016-07-12

omerblecher:


12_listcomprehension_lab/01.py
```diff
@@ -1,8 +1,5 @@
...
-Hint: look for chr() and ord()
-"""
-
-
-
+#List of all lower case letters
+
+#Generate a list of all lower letters
+lowerletters = [chr(index) for index in range(1,256) if index >= ord('a') and index <= ord('z')]
+print " ".join(lowerletters)
```
szabgab, line 5: It would be better to have the range from ord('a') till ord('z') and then there is no need for the 'if' statement.

12_listcomprehension_lab/02.py
```diff
@@ -1,6 +1,10 @@
...
+#All three letters permutation
+
+#Generate a list of all lower letters
+lowerletters = [chr(index) for index in range(1,256) if index >= ord('a') and index <= ord('z')]
+
+#Generate all three letters permutations
+threeletters = [a + b + c for a in lowerletters for b in lowerletters for c in lowerletters]
+
+#Print all permutations
+print " ".join(threeletters)
```
szabgab, line 10: Same comment as in 01.

omerblecher, line 10: Done (both 01 & 02)

12_listcomprehension_lab/01.py
omerblecher, line 5: Done

12_listcomprehension_lab/01.py
```diff
@@ -1,5 +1,5 @@
 #List of all lower case letters
 
 #Generate a list of all lower letters
-lowerletters = [chr(index) for index in range(1,256) if index >= ord('a') and index <= ord('z')]
+lowerletters = [chr(index) for index in range(ord('a'),ord('z'))]
 print " ".join(lowerletters)
```
szabgab, line 5: Please note, it should be `range(ord('a'), ord('z')+1)` in order to include 'z' as well.

PR #111, opened 2016-07-13

konst1988:


10_datastructures_lab/01.py
```diff
@@ -1,10 +1,15 @@
...
-    apple => red
-    lettuce => green
-    lemon => yellow
-    orange => orange
-"""
+import sys
+hosts = {}
+
+#read all hosts from file
+with open('4.Collections\\hosts', 'r') as f:
```
ynonp, line 5: 1. Better to take the file name from command line
2. Prefer `os.path.join` to concatenate file path parts such as: `os.path.join('4.collections', 'hosts')`

10_datastructures_lab/01.py
```diff
@@ -1,10 +1,15 @@
...
+    for line in f:
+        keyvalue = line.split("=")
```
ynonp, line 7: How about: `name, ip = line.rstrip().split('=')`
Now you don't need the indexes in the next line and everything's easier to read

10_datastructures_lab/02.py
```diff
@@ -1,5 +1,21 @@
...
+       found = False
+       for anagram in anagrams:
+           if anagram[0] == wordCount :
+               found = True
+               print "{} {}".format(word, anagram[1])
+
+        # Add to list if not found
+       if not found:
+            anagrams.append((wordCount, word))
+
```
ynonp, line 21: Python has an interesting construct to deal with such cases called the for...else. You can read about it here:
http://psung.blogspot.co.il/2007/12/for-else-in-python.html

PR #112, opened 2016-07-13

Guyemin:


12_listcomprehension_lab/01.py
```diff
@@ -1,8 +1,6 @@
...
-"""
-
-
-
+import sys 
+import os
+array = []
+list_ascii = range(97, 123)
+for i in list_ascii:
+	print chr(i)
```
szabgab, line 6: 👍

12_listcomprehension_lab/02.py
```diff
@@ -1,6 +1,10 @@
...
+import sys 
+import os
+
+a = range (97,123)
+b = range (97,123)
+c = range (97,123)
+
+letters = (range(97,123))
+match = [chr(x)+chr(y)+chr(z) for x in letters for y in letters for z in letters]
+print match
```
szabgab, line 10: - a,b,c seem to be unused here.
- There is no need for the parentheses around range in line 8

PR #113, opened 2016-07-13

tomag:


10_datastructures_lab/01.py
```diff
@@ -1,10 +1,34 @@
...
+entries = {}
+
+if (len(sys.argv) != 2):
+    exit("Usage: 01.py <hostsFile>")
+
+(fileName) = sys.argv[1]
+
+with open(fileName, "r") as hosts:
+    for line in hosts:
+        (hostName, ip) = line.split("=")
```
ynonp, line 18: Python's happy without the parenthesis - `hostname, ip = ...`

10_datastructures_lab/01.py
```diff
@@ -1,10 +1,34 @@
...
+and lets the user choose a host and get its ip address.
+If the host is not in the list, a message is shown to the user
+"""
+
+import sys
+
+entries = {}
+
+if (len(sys.argv) != 2):
+    exit("Usage: 01.py <hostsFile>")
```
ynonp, line 12: prefer `sys.exit()` over `exit`. There's a detailed discussion and explanation on their differences here:
http://stackoverflow.com/questions/6501121/the-difference-between-exit-and-sys-exit-in-python

10_datastructures_lab/01.py
```diff
@@ -1,10 +1,34 @@
...
+
+(fileName) = sys.argv[1]
```
ynonp, line 14: Parenthesis is not required here. Prefer `filename = sys.argv[1]`

10_datastructures_lab/02.py
```diff
@@ -1,5 +1,28 @@
...
+
+import sys
+from collections import Counter
+
+if (len(sys.argv) != 2):
+    exit("Usage: 02.py <fileName>")
+
+fileName = sys.argv[1]
+
+repeats = {}
```
ynonp, line 14: A defaultdict can work better here. If `repeats = defaultdict(list)`. With that in place you won't need the if/else because values not yet created will automatically return a new empty array

PR #114, opened 2016-07-13

razavishay:


10_datastructures_lab/01.py
```diff
@@ -1,10 +1,19 @@
...
-    lemon => yellow
-    orange => orange
-"""
+import sys
+names=sys.argv[1:]
+print "names=", names
+hostsFile = "C:\\Python Course\\2\\hosts.txt"
+hostsDict = {}
+with open(hostsFile, 'r') as f:
+    for line in f:
```
ynonp, line 7: I prefer the more simple approach:
`host, ip = line.rstrip().split('=')`

You get both variables, and rstrip removes the newline character from the end

10_datastructures_lab/02.py
```diff
@@ -1,5 +1,26 @@
...
+dict={}
+with open(wordsFile, 'r') as f:
+    for line in f:
+        word=line.strip()
+        words.append(word)
+        cnt=Counter(word)
+        if cnt not in counterList:
+            counterList.append(cnt) 
+
+for cnt in counterList:
```
ynonp, line 16: This works but can take a long time if the file is large. For example try to find anagrams in:
https://users.cs.duke.edu/~ola/ap/linuxwords

See if you can find a faster approach that doesn't require a nested loop to print everything

PR #115, opened 2016-07-13

tomag:


12_listcomprehension_lab/01.py
```diff
@@ -1,8 +1,10 @@
...
-Hint: look for chr() and ord()
-"""
-
-
-
+"""
+The program prints all english small letters 
+"""
+
+startIndex = 97
```
ynonp, line 5: `start_index = ord('a')` is easier to understand

PR #117, opened 2016-07-13

mashaGC:


10_datastructures_lab/01.py
```diff
@@ -1,10 +1,20 @@
...
+        host = line.split('=')[0] 
+        ip = line.rstrip('\n').split('=')[1]
+        hostsDic[host] = ip
+
+for host_input in sys.argv[1:]:
+    value = hostsDic.get(host_input,0)
+    if value == 0:
+        print "host:'%s' doesn't exist" % host_input
+    else:
+        print value
```
szabgab: nice
This is a more readable version of it:

```
        host, ip = line.rstrip('\n').split('=')
        hostsDic[host] = ip
```

Also it is clearer and more Pythonic to write

`if host_input in hostDic:` instead of lines 16-17

Lastly, please use .format instead of % in line 18.

10_datastructures_lab/01.py
```diff
@@ -8,13 +8,11 @@
...
 
 for host_input in sys.argv[1:]:
-    value = hostsDic.get(host_input,0)
-    if value == 0:
-        print "host:'%s' doesn't exist" % host_input
+    if host_input in hostsDic:
+        print hostsDic.get(host_input)
     else:
-        print value
+        print "host:'{}' doesn't exist".format(host_input)
```
szabgab, line 18: 👍

10_datastructures_lab/02.py
```diff
@@ -1,5 +1,35 @@
...
+    delList = []
+    print("")
+    word = wordsDic.keys()[0]
+    counter_i = wordsDic.get(word)
+    sys.stdout.write(word+" ")
+    del wordsDic[word]
+    for word_i in wordsDic:
+        if  counter_i == wordsDic.get(word_i):
+            sys.stdout.write(word_i+" ")
+            delList.append(word_i)
```
szabgab: This seems to be a bit too complex and as I can see it runs O(n^2)
You could make it simpler and faster by using a dictionary where the keys are the sorted version of each string and the values are the original strings. Each key that has more than one entry at the end has a list of anagrams.

10_datastructures_lab/02.py
```diff
@@ -13,23 +13,17 @@
...
-    del wordsDic[word]
-    for word_i in wordsDic:
-        if  counter_i == wordsDic.get(word_i):
-            sys.stdout.write(word_i+" ")
-            delList.append(word_i)
\ No newline at end of file
+for k,v in wordsDic.items():
+    print ""
+    for i in v:
+        sys.stdout.write(i+"\t")
```
szabgab, line 29: 👍

PR #118, opened 2016-07-13

razavishay:


12_listcomprehension_lab/01.py
```diff
@@ -1,8 +1,2 @@
-"""
-Use range() and list comprehension to get
-the list of all lowercase english letters
-Hint: look for chr() and ord()
-"""
-
-
-
+for char in range(97,97+26):
```
ynonp, line 1: Use list comprehensions:
https://www.tocode.co.il/bundles/python-verint/lessons/list-comprehension?tab=article

PR #119, opened 2016-07-13

konst1988:


12_listcomprehension_lab/01.py
```diff
@@ -4,5 +4,7 @@
 Hint: look for chr() and ord()
 """
 
-
+numbers = range(97, 97+26)
```
ynonp, line 7: `range(ord('a'), ord('z'))` is easier to understand

PR #120, opened 2016-07-13
ToCode Solution: תרגול קבצים

shlomitTal:


08_files_lab/01.py
```diff
@@ -1,6 +1,16 @@
...
+#D:\tmp\t1.txt D:\tmp\t2.txt
+
+import sys
+if len(sys.argv) < 3:
+    print "error -not enugh args"
+    sys.exit
+
+file1 = sys.argv[1]
+file2 = sys.argv[2]
+
```
ynonp, line 11: Prefer `file1, file2 = sys.argv[1:]`

08_files_lab/01.py
```diff
@@ -1,6 +1,16 @@
...
-the end of the first one
-"""
-
+#Commited by Shlomit
+#D:\tmp\t1.txt D:\tmp\t2.txt
+
+import sys
+if len(sys.argv) < 3:
+    print "error -not enugh args"
+    sys.exit
```
ynonp, line 7: 1. Line 7 does nothing. To actually call sys.exit you need to write parenthesis after function name such as `sys.exit()`
2. Better to write the error message inside `sys.exit()`. It prints the error to stderr AND returns an error code to the operating system.

08_files_lab/02.py
```diff
@@ -1,7 +1,17 @@
...
+
+with open(sys.argv[1], "r") as fr1:
+    with open(sys.argv[2], "r") as fr2:
+        with open(sys.argv[3], "w") as fw3:
+            for rows in itertools.izip_longest(fr1, fr2):
+                for i in rows:
+                    if i is not None:
+                     fw3.write(i)
+
+
```
ynonp, line 17: 👍

PR #121, opened 2016-07-13

shlomitTal:


10_datastructures_lab/01.py
```diff
@@ -1,10 +1,23 @@
...
+    sys.exit
+
+hashKey={}
+file = sys.argv[1]
+with open(file, "r") as f:
+    for line in f:
+        fields = line.split("=")
+        if len(fields) != 2:
+            continue
+        hashKey[fields[0]] = fields[1]
```
ynonp, line 15: It'd be cleaner to first extract the values into variables:

```
host, ip = fields
hashKey[host] = ip
```

10_datastructures_lab/02.py
```diff
@@ -1,5 +1,19 @@
...
+
+if len(sys.argv) < 2:
+    print "error - you didn't add any arguments!"
+    sys.exit
+file = sys.argv[1]
+hashmap=defaultdict(str)
+
+with open(file, "r") as f:
+    for row in f:
+       rowVal = row[:-1] if row.endswith("\n") else row
```
ynonp, line 13: Python has rstrip(): `rowVal = row.rstrip()` (which does the same thing)

PR #122, opened 2016-07-13

dorsegal:


10_datastructures_lab/01.py
```diff
@@ -1,10 +1,16 @@
...
+       (name, ip) = line.split('=')
+       hostsfile[name] = ip
+
+
+for name in sys.argv[1:]:
+    try:
+        print "%s = %s" % (name, hostsfile[name])
+
+    except:
+        print "%s does not exist\n" % name
```
szabgab: Nice, but it would be cleaner to check if the name is in the hostsfile instead of expecting an exception:
`if name in hostsfile:` will do it.

In addition, it would be better to use .format instead of % to format the output strings.

10_datastructures_lab/01.py
```diff
@@ -4,13 +4,11 @@
...
-    try:
-        print "%s = %s" % (name, hostsfile[name])
-
-    except:
-        print "%s does not exist\n" % name
\ No newline at end of file
+    if name in hostsfile:
+        print "{0} IP is {1}".format(name, hostsfile[name])
+    else:
+        print "{0} does not exist".format(name)
```
szabgab, line 14: 👍

10_datastructures_lab/02.py
```diff
@@ -1,5 +1,12 @@
...
+anagrams = defaultdict(str)
+
+with open("file.txt" ,"r") as fin:
+    for line in fin:
+        word = line.rstrip("\n")
+        key = ''.join(sorted(word.lower()))
+        anagrams[key] += word + " "
+
+for item in anagrams:
+    print anagrams[item]
```
szabgab, line 12: 👍

PR #124, opened 2016-07-14

nirna:


10_datastructures_lab/01.py
```diff
@@ -1,10 +1,15 @@
...
+""" Return IPs from hosts file per input machine name(s) """
+import sys
+from collections import defaultdict
+# Read input args
+host_file_path, pc_names = sys.argv[1], sorted(sys.argv[2:])
+# Read hosts into a dictionary
+dict_hosts = defaultdict(lambda: "No IP")
+with open(host_file_path,'r') as fin:
+    for line in fin: 
+        (host_name, host_ip) = line.rstrip().split("=")
```
ynonp, line 10: No need for parenthesis

10_datastructures_lab/02.py
```diff
@@ -1,5 +1,20 @@
...
+anagrams_file_path = sys.argv[1]
+# Read words into a dictionary holding their histogram as key
+dict_anagrams = defaultdict(list)
+with open(anagrams_file_path,'r') as fin:
+    for word in fin: 
+        # Generate a sorted lower case letters histogram
+        word = word.rstrip().lower()
+        sorted_dict_letters = ""
+        dict_letters = Counter(word)
+        for k in sorted(dict_letters): sorted_dict_letters += k + str(dict_letters[k])
```
ynonp, line 15: 1. You don't need to both count and sort - sorting the word's letters is enough. Consider "add" and "dad", when sorted both yield "add"
2. A sorted(str) can be used as the key in your dictionary, saving you the loop at line 15

PR #125, opened 2016-07-14

zfigov:


10_datastructures_lab/01.py
```diff
@@ -1,10 +1,27 @@
...
+Program should print the IP addresses of the hosts requested
 """
+import sys
+from itertools import izip
+
+(_,file1) = sys.argv
+
+host_ip = {}
+with open(file1) as fin:
+    a = [(line.rstrip('\n')).split("=") for line in fin]
```
ynonp, line 15: Better to use a single loop to read the file:

```
open ...:
  for line in file:
    host, ip = line.rstrip().split('=')
    host_ip[host] = ip
```

10_datastructures_lab/02.py
```diff
@@ -1,5 +1,43 @@
...
+
+# count the number of letters in each word
+words_count = []
+for word in words:
+    w = Counter(word)
+    words_count.append(w)
+
+# compare each word to the other using the counter
+found = [False for i in range(len(words))]
+
```
ynonp, line 30: It should be more efficient to build a dictionary. The key is the sorted letters of a word and the value is the list of words keyed that way. So for the example input you'll end up with:

```
'add' => 'add', 'dad'
'emor' => 'more', 'rome',
```

PR #126, opened 2016-07-14

razavishay:


14_functions_lab/01.py
```diff
@@ -1,7 +1,14 @@
...
-
+def SumNumbers (list):
+    s = 0
+    for i in list:
+        if type(i) == int:
+            s += i
+    return s
+
+def MultiplyNumbers (list):
+    m=1
```
ynonp, line 9: List Comprehension + reduce can yield a simpler solution:

```
import operator
return reduce operator.mul, [n for n in list if isinstance(n, numbers.Number)]
```

14_functions_lab/02.py
```diff
@@ -1,7 +1,7 @@
-"""
-Write a function that takes two arguments: 
-    A string
-    And a number
-If wrong types were passed in, raise an exception
-"""
-
+def checktypes (num, string):
+    if type(num) <> int:
```
ynonp, line 2: Use `!=` to say not equals than in python

14_functions_lab/03.py
```diff
@@ -1,5 +1,14 @@
-"""
-Write a function that calculates the sum
-of the 10th digit from all arguments passed to it
-"""
-
+def TensSum(*list):
```
ynonp, line 1: Please leave the top comments in the file (the ones inside """). They make it easier for everyone opening the file later to understand what's inside

14_functions_lab/03.py
```diff
@@ -1,5 +1,14 @@
-"""
-Write a function that calculates the sum
-of the 10th digit from all arguments passed to it
-"""
-
+def TensSum(*list):
+    s = 0
+    for number in list:
+        if type(number) == int:
```
ynonp, line 4: Leaving it numeric yields a simpler solution: `(number / 10) % 10` is exactly the digit you need

14_functions_lab/04.py
```diff
@@ -1,6 +1,8 @@
-"""
-Write a function that takes minlen and
-a list of words, and returns only the words
-longer than minlen
-"""
-
+def longer_than(minLength, *list):
+    refinedList = []
+    for word in list:
```
ynonp, line 3: Notice how github colours "list" in the same color as print (and differently from "refinedList" and "word"). The reason is python has a function named list, so when you use that as a variable name it can be a bit confusing.

14_functions_lab/04.py
```diff
@@ -1,6 +1,8 @@
...
-Write a function that takes minlen and
-a list of words, and returns only the words
-longer than minlen
-"""
-
+def longer_than(minLength, *list):
+    refinedList = []
+    for word in list:
+        if len(word) > minLength:
+            refinedList.append(word)
```
ynonp, line 5: That's why they invented list comprehension :). Consider:

```
return [word for word in list if len(word) > minlen]
```

14_functions_lab/04.py
```diff
@@ -1,6 +1,8 @@
-"""
-Write a function that takes minlen and
-a list of words, and returns only the words
-longer than minlen
-"""
-
+def longer_than(minLength, *list):
```
ynonp, line 1: Use lowercase letter only for variable names

14_functions_lab/05.py
```diff
@@ -1,10 +1,12 @@
...
-Write a function groupby that takes a list
-and a function and returns a dictionary
-keyd by the return value of the function on the list items
-
-For example:
-    groupby(lambda s: s[0], ['foo', 'fi', 'hello', 'hi'])
-    returns: { 'f': ['foo','fi'], 'h': ['hello', 'hi'] }
-"""
-
+def GroupBy(func, myList):
```
ynonp, line 1: Use lowercase only function names

14_functions_lab/05.py
```diff
@@ -1,10 +1,12 @@
...
+    myDict = {}
+    for el in myList:
+        theKey =  func(el)
+        if myDict.get(theKey) == None:
```
ynonp, line 5: get has a second argument, which is the value to return if key is not found. So better to use: `myDict.get(theKey, []).append(el)`

PR #127, opened 2016-07-14

zfigov:


12_listcomprehension_lab/01.py
```diff
@@ -4,5 +4,5 @@
 Hint: look for chr() and ord()
 """
 
-
+lower_keys = [chr(lc) for lc in range(255) if (chr(lc)).islower()]
```
ynonp, line 7: Better to use `ord('a')` and `ord('z')` as the limits:

```
[chr(x) for x in range(ord('a'), ord('z')+1)]
```

PR #128, opened 2016-07-14

zfigov:


14_functions_lab/01.py
```diff
@@ -4,4 +4,30 @@
     mymul - returns the multiplication of its input arguments
     Ignore non-numeric arguments
 """
+import operator 
 
+def mysum(*paramsin):
+    res = 0
```
ynonp, line 10: Why not simply `return sum([int(val) for val in paramsin if type(val) is int])` ?

ynonp, line 10: A better way to check for numeric value is `isinstance(n, numbers.Number)`. That also checks for floats

14_functions_lab/01.py
```diff
@@ -4,4 +4,30 @@
...
+    res = sum([int(val) for val in paramsin if type(val) is int])
+    
+    return res
+
+def mymul(*paramsin):   
+    if len(paramsin)>0:
```
ynonp, line 16: Since *1 is a noop, you can just append a fake 1 to the end of the list:

```
reduce(operator.mul,[int(val) for val in paramsin + [1] if type(val) is int])
```

14_functions_lab/03.py
```diff
@@ -3,3 +3,9 @@
 of the 10th digit from all arguments passed to it
 """
 
+def sum_tens_digit(*numbers):
+
+    val = sum([int(str(n)[-2]) for n in numbers])
```
ynonp: Solution breaks if any value is shorter than 3 digits. Can you fix it?

14_functions_lab/01.py
zfigov, line 10: It's usually easier to debug when the return value is seperate.

ynonp:
Great work. Closing.

PR #129, opened 2016-07-14

konst1988:


14_functions_lab/01.py
```diff
@@ -4,4 +4,20 @@
     mymul - returns the multiplication of its input arguments
     Ignore non-numeric arguments
 """
+import operator
 
+def perform(action, *nums):
```
ynonp, line 9: Nice! Bonus 5 points ⭐

14_functions_lab/01.py
```diff
@@ -4,4 +4,20 @@
...
+    return reduce(action, [x for x in list(nums[0]) if isinstance(x, int)])
+
+def mysum(*nums):
+    return perform(operator.add, nums)
+
+def mymul(*nums):
+    # add 1 just in case list is empty
+    newNums = list(nums)
+    newNums.append(1)
+    return perform(operator.mul, newNums)
```
ynonp, line 19: Or even better: `return perform(operator.mul, nums + [1])`

14_functions_lab/03.py
```diff
@@ -3,3 +3,7 @@
 of the 10th digit from all arguments passed to it
 """
 
+def sum10th(*nums):
+    return sum([int(str(x)[-2]) for x in nums])
```
ynonp, line 7: Solution breaks on inputs that are shorter than 3 letters. Can you fix it?

14_functions_lab/05.py
```diff
@@ -8,3 +8,13 @@
...
+from collections import defaultdict
+
+def groupby(func, lst):
+    result = defaultdict(list)
+    for item in lst:
+        key = func(item)
+        result[key].append(item)
+    return result
+
+print groupby(lambda s: s[0], ['foo', 'fi', 'hello', 'hi'])
```
ynonp, line 20: 👍

PR #130, opened 2016-07-14

shlomitTal:


12_listcomprehension_lab/01.py
```diff
@@ -1,8 +1,3 @@
-"""
-Use range() and list comprehension to get
-the list of all lowercase english letters
-Hint: look for chr() and ord()
-"""
-
-
-
+#commited by Shlomit
+list = [chr(i) for i in range (97, 123)]
```
ynonp, line 2: Better to use `ord('a')` and `ord('z') + 1` as limits

PR #132, opened 2016-07-14

tomag:


14_functions_lab/01.py
```diff
@@ -1,7 +1,21 @@
...
-    Ignore non-numeric arguments
-"""
-
+"""
+"""
+
+def mysum(*numbers):
+    sum = 0
+    for number in numbers:
+        if type(number) is int:
```
ynonp, line 7: A better way to check if something is a number is:

```
isinstance(n, numbers.Number)
```

This also works for floats

14_functions_lab/02.py
```diff
@@ -1,7 +1,15 @@
...
+import sys
+
+def isrighttype(number, text):
+    if number.isdigit() == False:
+        raise Exception("Input '{}' is not a valid number".format(number))
+
+if len(sys.argv) != 3:
+    sys.exit("Usage: 02.py <number> <string>")
+
+(number, text) = sys.argv[1:]
```
ynonp, line 13: No need for parenthesis: `number, text = sys.argv[1:]` works well

14_functions_lab/04.py
```diff
@@ -1,6 +1,11 @@
...
-longer than minlen
-"""
-
+"""
+"""
+
+def longer_than(length, *text):
+    for word in text:
+        if len(word) > length:
+            yield word
```
ynonp, line 7: Note that the task was to return a list. Yield works for printing but if we wanted for instance the length yield wouldn't be that great. The following code breaks in your program:

```
print len(longer_than(3, "tomer", "nir", "hello", "a"))
```

PR #136, opened 2016-07-14

Oransha:


12_listcomprehension_lab/01.py
```diff
@@ -1,8 +1,5 @@
...
-Hint: look for chr() and ord()
-"""
-
-
-
+lst = []
+ascii = range(97, 123)
+for i in ascii:
+     lst.append([ x for x in chr(i)])
+print lst
```
szabgab, line 5: That's not really a list comprehension. Try getting rid of the 'for' loop and move the whole thing inside the [].
Also the variable name ascii isn't really correct  all the characters between 0-127 are ascii. What we have here is maybe lower_case_letters. or ascii_of_lower_case_letter.

12_listcomprehension_lab/02.py
```diff
@@ -1,6 +1,14 @@
...
+lst3 = []
+ascii = range(97, 123)
+for i in ascii:
+     lst1.append([ x for x in chr(i)])
+     lst2.append([ x for x in chr(i)])
+     lst3.append([ x for x in chr(i)])
+
+sequence = [ (x + y + z) for x in lst1 for y in lst2 for z in lst3 ]
+
+print sequence
```
szabgab, line 14: The sequence is now generated with a real list comprehension, but the letters themselves not yet.
Also, you don't need 3 lst\* variables. One will do.

12_listcomprehension_lab/01.py
Oransha, line 5: ascii_for_lower_case_letters = range(97, 123)
lst = [chr(i) for i in ascii_for_lower_case_letters]
print lst

12_listcomprehension_lab/02.py
Oransha, line 14: import operator

ascii_for_lower_case_letters = range(97, 123)

char = [chr(i) for i in ascii_for_lower_case_letters]

sequence = [ (x+y+z) for x in char for y in char for z in char ]

print sequence

szabgab, line 14: 👍

PR #137, opened 2016-07-14

dorsegal:


12_listcomprehension_lab/01.py
```diff
@@ -4,5 +4,7 @@
 Hint: look for chr() and ord()
 """
 
+lowletters = [chr(l) for l in range(255) if chr(l).islower()]
 
+print lowletters
```
szabgab, line 10: Maybe that islower protects you here, but it would be much better to have the range run only on the characters you'd like to print and not all the ascii characters.

BTW Have you tried printing the characters under 32 ascii? It is spectacular.

12_listcomprehension_lab/02.py
```diff
@@ -1,6 +1,13 @@
...
 """
+from collections import defaultdict
+import itertools
 
+anagrams = defaultdict(str)
+
+lowletters = [chr(l) for l in range(255) if chr(l).islower()]
+
+for letter in itertools.combinations_with_replacement(lowletters, 3): 
+        print(''.join(letter))
```
szabgab: The same comment to be more explicit about the range.

12_listcomprehension_lab/01.py
dorsegal, line 10: I know :) i used to program in assembly

szabgab, line 10: 👍

PR #138, opened 2016-07-14

moshekenan:


14_functions_lab/01.py
```diff
@@ -1,7 +1,26 @@
...
-    Ignore non-numeric arguments
-"""
-
+'''
+functions no.1
+'''
+def func_sum(*params):
+    res = 0
+    for param in params:
+        if type(param) == int:
```
ynonp, line 7: A better way to check isinstance(x, numbers.Number), because it also works for floats:

```
import numbers
print isinstance(10, numbers.Number)
print isinstance(10.5, numbers.Number)
print isinstance('bye', numbers.Number)
```

14_functions_lab/04.py
```diff
@@ -1,6 +1,14 @@
-"""
-Write a function that takes minlen and
-a list of words, and returns only the words
-longer than minlen
-"""
-
+'''
+functions no4
+'''
+def longer_than(long, *words):
```
ynonp, line 4: Can also use list comprehension:

```
return [word for words if len(word) > long]
```

PR #141, opened 2016-07-14

lahavz:


14_functions_lab/01.py
```diff
@@ -1,7 +1,26 @@
...
-
+"""
+Two fucntions: 1st is sum , 2nd is multiply.
+"""
+
+from types import *
+
+def sum(*args):
+    res = 0
+    for x in args:
```
ynonp, line 9: Better to use numbers.Number to check for numeric value, as that also covers floats:

```
import numbers

# True
print isinstance(10, numbers.Number)
# True
print isinstance(10.5, numbers.Number)

# False
print isinstance('bye', numbers.Number)
```

14_functions_lab/04.py
```diff
@@ -1,6 +1,20 @@
...
-
+"""
+function return words greather than length
+"""
+from types import *
+
+
+
+def func(length,*args):
+    res = []
```
ynonp, line 9: That's why they invented list comprehensions :)

```
return [word for word in args if len(word) > length]
```

PR #145, opened 2016-07-14

shaiMar:


12_listcomprehension_lab/01.py
```diff
@@ -4,5 +4,7 @@
 Hint: look for chr() and ord()
 """
 
-
-
+oa = ord('a')
+oz = ord('z')
+list = [chr(X) for X in range(256) if X>=oa and X<=oz]
```
ynonp, line 9: Better to use `ord('a')` and `ord('z')+1` as arguments to range

PR #146, opened 2016-07-14

nirna:


14_functions_lab/01.py
```diff
@@ -1,7 +1,23 @@
...
-"""
-
+""" func1: sum all inputs; func2: multiply all inputs """
+import numbers
+
+def mysum(*nums):
+    return sum([n for n in nums if isinstance(n, numbers.Number)])
+
+def mymul(*nums):
+    mul = 1
```
ynonp, line 8: Python still has `reduce` so you can:

```
import operator
numbers_only = [n for n in nums if isinstance(n, numbers.Number)]
reduce operator.mul, numbers_only
```

ynonp:
Good Work. Closing.

PR #152, opened 2016-07-16

eladkah:


10_datastructures_lab/01.py
```diff
@@ -1,10 +1,26 @@
...
+if sys.argv == 1:
+    print 'Usage: %s [Machine Names...]' % sys.argv[0]
+
+wanted_machines = sys.argv[1:]
+exist_machines = {}
+
+# Read file
+with open('hosts.txt', 'r') as f:
+    for line in f:
+        splitted = line.split('=')
```
ynonp, line 19: Better to separate to variables and `rstrip()` with: `host, ip = line.rstrip().split('=')`

10_datastructures_lab/01.py
```diff
@@ -1,10 +1,26 @@
...
+        exist_machines[splitted[0]] = splitted[1].rstrip('\n')
+
+for machine in wanted_machines:
+    if exist_machines.has_key(machine):
```
ynonp, line 23: For compatibility with Python3, it's better to use `if machine in exist_machines` (which does the same as `has_key` but works in python3)

10_datastructures_lab/02.py
```diff
@@ -1,5 +1,27 @@
...
+Prints anagrams in same line
+Usage: 02.py <file_path>
+'''
+
+import sys
+import os
+if len(sys.argv) != 2 or not os.path.exists(sys.argv[1]):
+    print "Usage: %s <file path>" % sys.argv[0]
+
+dictionary = {}
```
ynonp, line 12: A defaultdict would save you the if later on. Define with `dictionary = collections.defaultdict(list)` and then fetching a non existent value yields a new empty list so it's always safe to `dictionary[key].append(word)`

PR #153, opened 2016-07-16

eladkah:


14_functions_lab/01.py
```diff
@@ -1,7 +1,19 @@
...
-"""
-
+'''
+mysum calcs sum of parameters while ignoring non-int items
+mymul calcs mul of parameters while ignoring non-int items
+'''
+
+
+def mysum(*n):
+    return sum(i for i in n if type(i) is int)
```
ynonp, line 8: numbers.Number is a better indicator for numeric value, since it also covers floats. Here's how to use it:

```
import numbers
def mysum(*n):
  return sum(i for i in n if isinstance(i, numbers.Number))
```

14_functions_lab/01.py
```diff
@@ -1,7 +1,19 @@
...
+
+
+def mymul(*n):
+    mul = 1
```
ynonp, line 12: reduce is fantastic in such situations: `reduce(operator.mul, numbers)`

14_functions_lab/05.py
```diff
@@ -1,10 +1,16 @@
...
-    groupby(lambda s: s[0], ['foo', 'fi', 'hello', 'hi'])
-    returns: { 'f': ['foo','fi'], 'h': ['hello', 'hi'] }
-"""
-
+'''
+returns dictionary where key is result of function and value is list of strings that returned the same result
+'''
+
+def group_by(s, *words):
+    dictionary = {}
```
ynonp, line 6: With a defaultdict you don't need the if later on. This code works:

```
import collections
d = collections.defaultdict(list)
d['x'].append('y')
```

PR #155, opened 2016-07-16

HananRosenthal:


14_functions_lab/04.py
```diff
@@ -1,6 +1,9 @@
...
-
+
+def get_words(max_len, *words):
+    list=[]
+    for word in words:
+        if len(word) > max_len:
+            list.append(word)
+    return  list
+
+print( get_words(3,'a','aa','vvdd', 'tresd', 'ww') )
```
szabgab: A list comprehension might be better here.

14_functions_lab/05.py
```diff
@@ -1,10 +1,16 @@
...
+    myDict = {}
+    list=[]
+    for c in string.ascii_letters:
+        list = f(c, words)
+        if len(list) > 0:
+            myDict[c] = list
+    print(myDict)
+
+words = ["hello", "hi", "help", "bye", "here", "modify", "movie", "ticho", "time"]
+groupby(f, *words)
```
szabgab, line 16: This is not exactly what the exercise was. In this solution the function (f) passed to the groupby function expects two parameters and there is some internal data injection with the for loop on line 9. Can you change your solution so that there won't be a need for ascii_letters and that f would accept only a single paramteer?

14_functions_lab/04.py
```diff
@@ -1,9 +1,5 @@
...
-    list=[]
-    for word in words:
-        if len(word) > max_len:
-            list.append(word)
-    return  list
+    return [word for word in words if len(word) > max_len]
 
-print( get_words(3,'a','aa','vvdd', 'tresd', 'ww') )
+	
+print( get_words(4,'a','aa','vvdd', 'tresd', 'ww') )
```
szabgab, line 5: 👍 nice

14_functions_lab/05.py
HananRosenthal, line 16: maybe this one?

import collections
import string

f = lambda wList: [w[0] for w in wList]

def groupby(f, *words):
    myDict = {}
    key = f(words)
    for k in key:
        myDict[k] = [word for word in words if word[0] == k]
    print(myDict)

words = ["hello", "hi", "help", "bye", "here", "modify", "movie", "ticho", "time"]
groupby(f, *words)

14_functions_lab/04.py
HananRosenthal, line 5: The function is:

def get_words(max_len, *words):
    return [word for word in words if len(word) > max_len]

14_functions_lab/05.py
szabgab, line 16: Nope. Here you call f with the whole list. You need to call f with each one of the elements of the 'words' list.

HananRosenthal:
hello Gabor,

need some help with 05.py...
10X

szabgab:
What kind of help?

HananRosenthal:
1. what exactly the lamda function must get, and what should it perform
2. what should the main function perform
   10X

HananRosenthal:
i mean, if lambda only gets a word and return its first letter - its stupid

szabgab:
Oh, Why do you feel it stupid?

HananRosenthal:
becuase this can be one code line in the "main" function, without the overhead of a stupid lambda function

szabgab:
This is a concept (of hashing) with a simple example.

What if someone will want to do the same grouping but based on the last character?
And then someone else might want to do the grouping with the sum of the ascii values of the strings. You would then need to copy, paste and adjust the grouping function.

With the solution of this exercise the implementor of the grouping function does not need to know about how the hashing algorithm works (first character, last character, sum of ascii, etc.), it can rely on the user supplying the hashing algorithm as the first function.

Let me give you another example: sorting. sorting has two major parts. One is what computer scientist usually call "sorting algorithm". This can be quick-sort, bubble-sort, merge-sort, etc.
The other one is what application programmers call the "sorting function" that given two values can decide which one is "bigger" and which one is "smaller". By de-coupling these two functionalities Python can provide a single "sorting algorithm" with a default "sorting function", but you as the user can easily replace the "sorting function" to be "compare the first character of the strings" or "compare the last character of the strings" or "compare the sum of the ascii numbers of the strings".

HananRosenthal:
ok, i got the idea.
last question here: how do i add different values (one at a time) to the same key in a dictionary?

szabgab:
For that each value (not the key) in the dict will have to be a list and then you can treat it as a list and append values to it.

PR #157, opened 2016-07-16

nirbar1985:


06_modules_lab/02.py
```diff
@@ -1,4 +1,21 @@
...
+write a program eivh gets in argv 2 numbers and prints its sum
+
+
+
+"""
+
+import sys
+
+if len(sys.argv) != 3:
+	sys.exit("please enter exactly 2 parameters")
```
ynonp, line 11: No need for this check. If length is different the program will fail later on and the next except will catch it (giving you the same result with a few less lines of code).

ynonp:
Fantastic. Closing.

PR #160, opened 2016-07-17

eladkah:


18_regexp_lab/01.py
```diff
@@ -1,12 +1,27 @@
...
+and a key (x) and prints the value of x (y)
+Uses regex to find relevant line
+'''
+
+import sys
+import os
+import re
+
+if len(sys.argv) != 3:
+    print 'Usage: %s <file_name> <key>' % sys.argv[0]
```
ynonp, line 15: Better to write the error message inside `sys.exit()`. It prints the error to stderr AND returns an error code to the operating system.

18_regexp_lab/01.py
```diff
@@ -1,12 +1,27 @@
...
+    sys.exit(1)
+
+if not os.path.exists(sys.argv[1]):
+    print 'File: %s does not exist' % sys.argv[1]
+    sys.exit(1)
+
+with open(sys.argv[1], 'r') as f:
+    for line in f:
+        result = re.search(r'%s[ ]*=[ ]*(\d*)' % sys.argv[2], line)
```
ynonp, line 24: Using external input as regexp is considered bad practice: The input may contain characters that have special meaning in regexp, such as `.`, `^`, `$`, etc. Better to read all values to a dictionary and search that.

PR #161, opened 2016-07-17

razavishay:


18_regexp_lab/01.py
```diff
@@ -1,12 +1,12 @@
...
-Each line in the file can either be:
-    An empty line
-    A comment line (Start with #)
-    A property line (of the form key = value)
-
-Write a program that takes a property file name and key
-as command line arguments and prints the requested value
-"""
-
+import sys
```
ynonp, line 1: Please leave the top comments in the file (the ones inside """). They make it easier for everyone opening the file later to understand what's inside

18_regexp_lab/03.py
```diff
@@ -1,15 +1,9 @@
...
-"""
+import re
+import sys
+
+filename = sys.argv[1]
+
+with open (filename, 'r') as f:
+    for line in f:
+        sys.stdout.write(re.sub(r'(\w+),(\w+),(.*)', lambda m: m.group(2)+ ',' + m.group(1)+ ',' + m.group(3), line))
+
```
ynonp, line 9: Good Work! As a bonus read about CSV module and try to solve the same exercise with that module:
https://docs.python.org/2/library/csv.html

PR #162, opened 2016-07-17

gildahan:


06_modules_lab/01.py
```diff
@@ -1,4 +1,6 @@
-""" Write a program that takes a count
-from sys.argv import and prints "Hello Python"
-count times
-"""
+import sys
+
+numOfPrints = int(sys.argv[1])
+
+for x in range(0,numOfPrints):
```
ynonp, line 5: You're not using the `x` nor the start value (0), so better to omit both. This works just as well:

```
for _ in range(numOfPrints):
```

06_modules_lab/01.py
```diff
@@ -1,4 +1,6 @@
-""" Write a program that takes a count
-from sys.argv import and prints "Hello Python"
-count times
-"""
+import sys
+
+numOfPrints = int(sys.argv[1])
```
ynonp, line 3: Better to use lowercase letters only for variable names, so num_of_prints is preferred.

06_modules_lab/02.py
```diff
@@ -1,4 +1,10 @@
-""" Write a program that reads 2 numbers from sys.argv
-and prints their sum.
-"""
-
+import sys
+import os
+
+if len(sys.argv) != 3 :
+    print "Not enogh paraeters "
```
ynonp, line 5: Better to write the error message inside `sys.exit()`. It prints the error to stderr AND returns an error code to the operating system.

06_modules_lab/03.py
```diff
@@ -1,11 +1,24 @@
...
+parser = argparse.ArgumentParser()
+parser.add_argument("Directory")
+parser.add_argument("FileSize")
+args = parser.parse_args()
+
+# (_,dir,fileSize)= sys.argv
+
+for root,dirs,files in os.walk(args.Directory):
+    for name in files:
+        fullFilePath = "{}\{}".format(root,name)
```
ynonp, line 14: There's `os.path.join` function which joins directory and name, taking care of the differences between unix and windows (unix path separator is /)

PR #163, opened 2016-07-17

dovron:


12_listcomprehension_lab/01.py
```diff
@@ -1,8 +1,13 @@
...
+      _      _      _
+   __(.)< __(.)> __(.)=
+   \___)  \___)  \___)   
+
+Python course - Verint
+Week 2
+
+"""
+
+word = [chr(i) for i in range(97,123)]
```
ynonp: Do you really remember that chr(97) == 'a'? Do you expect everyone reading your code to remember? Better to use ord('a') and ord('z') as the limits: `range(ord('a'), ord('z')+1)`

ynonp:
👍  Closing.

PR #164, opened 2016-07-17
ToCode Solution: תרגול קבצים

gildahan:


08_files_lab/02.py
```diff
@@ -1,7 +1,23 @@
...
+array1 = []
+array2 = []
+with open(fin1,"r") as f1:
+    for line in f1:
+        array1.append(line)
+with open(fin2,"r") as f2:
+    for line in f2:
+        array2.append(line)
+with open(fout,"a") as f3:
+    if len(array1)>len(array2):
```
ynonp, line 13: The function itertools.izip_longest provides a much nicer result. Its documentation is a bit scary, but here you can find a much better explanation with examples:
https://artemrudenko.wordpress.com/2013/04/11/python-zip-izip-and-izip_longest/

PR #166, opened 2016-07-17

dovron:


14_functions_lab/01.py
```diff
@@ -1,7 +1,23 @@
...
+   __(.)< __(.)> __(.)=
+   \___)  \___)  \___)   
+
+Python course - Verint
+Week 2
+
+"""
+
+def mysum(*arg):
+    res = 0
```
ynonp: You can also use Python's `sum()` function, by replacing the loop with list comprehension

14_functions_lab/04.py
```diff
@@ -1,6 +1,17 @@
...
+      _      _      _
+   __(.)< __(.)> __(.)=
+   \___)  \___)  \___)   
+
+Python course - Verint
+Week 2
+
+"""
+def longer_than(num,*words):
+    res = []
```
ynonp: List comprehension make this so much nicer:

```
return [word for word in words if len(word) > minlen]
```

PR #168, opened 2016-07-18

razavishay:


20_classes_lab/01.py
```diff
@@ -1,6 +1,9 @@
-"""
-The following code assumes a class
-named Summer exists. 
-Complete the class so correct result is printed
-"""
-
+class Summer(object):
+    def __init__(self, val = 0):
+        self.sum = val
+    def add(self, *vals):
```
ynonp, line 4: Use one blank line between methods to make the code much nicer to read. See here: https://www.python.org/dev/peps/pep-0008/#blank-lines

20_classes_lab/03.py
```diff
@@ -1,13 +1,33 @@
...
+        self.depends = []
+
+    def add_dependency(self, *dependencies):
+        for dependency in dependencies:
+            self.depends.append(dependency)
+
+    def build(self):
+        if self.depends != []:
+            for widget in self.depends:
+                print widget.name
```
ynonp, line 13: Not good enough: You should also print all widget's dependencies if they were not already printed

PR #170, opened 2016-07-18

HananRosenthal:


14_functions_lab/05.py
```diff
@@ -1,16 +1,15 @@
...
-            myDict[c] = list
+    for word in words:
+        k = f(word)
+        list = [w for w in words if w[0] == k]
+        myDict[k] = list
     print(myDict)
 
-words = ["hello", "hi", "help", "bye", "here", "modify", "movie", "ticho", "time"]
+words = ["hello", "hi", "help", "bye", "here", "modify", "movie", "typo", "time"]
 groupby(f, *words)
```
szabgab, line 12: This code still goes over the input O(n^2) once in the for loop and once in the list comprehension. You don't need the latter.

HananRosenthal, line 12: the later is required to compose a list of words begining with a certain letter

szabgab:
You already have the key in hand when you call `k = f(word)`. Use that as the key in the dict.

HananRosenthal:
yes, but still need to extract all words belonging to a specific key. How would you do that without another list comprehension?

szabgab:
You go over the words one by one and on each turn you add the word to the list of the respective key.
There is only one "loop" in this sentence.

14_functions_lab/05.py
```diff
@@ -1,15 +1,12 @@
...
     for word in words:
         k = f(word)
-        list = [w for w in words if w[0] == k]
-        myDict[k] = list
+        myDict.setdefault(k, []).append(word)
     print(myDict)
 
-words = ["hello", "hi", "help", "bye", "here", "modify", "movie", "typo", "time"]
+words = ["hello", "hi", "help", "bye", "here", "modify", "movie", "typo", "time", 'jump', 'jumbo', 'murder', 'jamboree']
 groupby(f, *words)
```
szabgab, line 12: 👍 yes, that's it!

HananRosenthal, line 12: sweated blood...

szabgab, line 12: Have a 🍪  you deserve it :)

PR #172, opened 2016-07-18

lahavz:


14_functions_lab/01.py
```diff
@@ -1,7 +1,26 @@
...
-"""
-
+"""
+Two fucntions: 1st is sum , 2nd is multiply.
+"""
+
+from types import *
+
+def sum(*args):
+    res = 0
```
ynonp, line 8: This works but has too much code duplication between sum and multiply. Can you reduce their duplicate code?

14_functions_lab/02.py
```diff
@@ -1,7 +1,37 @@
...
-
+"""
+function that gets two paramets and check validity. (str and number)
+"""
+from types import *
+
+
+
+def func(*args):
+    parameters = { 'integers' : 0, 'strings' : 0 }
```
ynonp, line 9: Nice, but that wasn't the question... The task was to enforce first parameter to be an int and second to be a string

14_functions_lab/03.py
```diff
@@ -1,5 +1,23 @@
...
+function sums the tenth numbers
+"""
+from types import *
+
+
+
+def func(*args):
+    res = 0
+    for x in args:
+        if (type(x) is int) and (int(x) >= 10): res += x / 10 % 10
```
ynonp, line 11: Notice how the main content of the function (to add the tens digit) is indented to the far right. This makes code harder to understand.

Can you modify the code so that part is in the center?

14_functions_lab/04.py
```diff
@@ -1,6 +1,20 @@
...
-
+"""
+function return words greather than length
+"""
+from types import *
+
+
+
+def func(length,*args):
+    res = []
```
ynonp, line 9: Note how in first glance it's not that obvious what the code returns. Using list comprehension I think you can re-order the code to make it clearer

PR #173, opened 2016-07-18

konst1988:


18_regexp_lab/01.py
```diff
@@ -9,4 +9,17 @@
...
+import re
 
+(_,filename,key) = sys.argv
+
+with open(filename, "r") as f:
+    for line in f:
+        if line[0] == '\n' or line[0] == '#': continue
+        keyvalue = re.search(r'^(\w+)[ ]*=[ ]*(\w+)$',line)
+        if keyvalue == None:
+            print "'{}' is not a key value line".format(line)
```
ynonp, line 22: Generally it's better to `continue` after this print, instead of writing most of the function code in the `elif` block

18_regexp_lab/03.py
```diff
@@ -13,3 +13,16 @@
...
 """
+
+import sys
+import re
+
+(_,filename) = sys.argv
+
+with open(filename, "r") as f:
+    for line in f:
+        fields = re.search(r'^(.*),(.*),(.*)$',line)
```
ynonp, line 24: Nice! Though I think you can simplify the code by using `sub` instead of `search`.

PR #174, opened 2016-07-18

moshekenan:


20_classes_lab/01.py
```diff
@@ -1,6 +1,31 @@
...
+import os
+import re
+
+class Summer(object):
+
+    def __init__(self):
+        self._total = 0
+
+    def add(self, *nums):
+        self._total += sum([n for n in nums])
```
ynonp, line 14: Looks like `sum(nums)` would do the same job with less code

20_classes_lab/03.py
```diff
@@ -1,13 +1,46 @@
...
-
+"""
+Classes and objects no 2
+"""
+import sys
+import os
+
+class Widget(object):
+
+    all_objects = []
```
ynonp, line 9: Keeping track of all built object is a waste. If an object should leave scope and needs to be deleted, memory won't be freed up because it has a copy in all_objects.

Also iterating an array of all the widgets ever built just to find out if a given widget was already built is a waste of CPU time.

Please try to find a better solution without the class variable

ynonp, line 9: And as a bonus, read about weakref and how they would solve some of the problems with `all_objects` class variable:
https://docs.python.org/2/library/weakref.html

(but keep in mind even with weakref - a class variable is still the wrong solution here)

PR #175, opened 2016-07-19

HananRosenthal:


18_regexp_lab/01.py
```diff
@@ -1,12 +1,14 @@
...
+parser = argparse.ArgumentParser(description='Returns a parameter that match a given key')
+parser.add_argument('-p','--path', help='definitions file', required = True)
+parser.add_argument('-k','--key', help='key', required = True)
+kuku = parser.parse_args()
+
+with open(kuku.path, 'r') as f:
+    for line in f:
+        if kuku.key in line:
+            res = re.search("=\s?(\w*)", line)
+            print(res.group(1))
```
szabgab: I don't think this will disregard a commented out property line like this:

```
# name = foo
```

neither lines where there are more than one spaces after the `=` sign.

18_regexp_lab/01.py
```diff
@@ -10,5 +10,7 @@
 with open(kuku.path, 'r') as f:
     for line in f:
         if kuku.key in line:
-            res = re.search("=\s?(\w*)", line)
+            if re.search('^\s?#', line):
+                continue
+            res = re.search("=\s+(\w*)", line)
             print(res.group(1))
```
szabgab: What about a comment line like `# bla?`  or `# name = foo` ?

HananRosenthal: any line begining with # shall be ignored - check it

szabgab: nope. If I have several spaces before the # then code and I run  with `-k name` then I get 'many'.

```
     # name = many spaces 2
```

18_regexp_lab/01.py
```diff
@@ -10,7 +10,7 @@
 with open(kuku.path, 'r') as f:
     for line in f:
         if kuku.key in line:
-            if re.search('^\s?#', line):
+            if re.search('^\s+#', line):
                 continue
             res = re.search("=\s+(\w*)", line)
             print(res.group(1))
```
szabgab: Better, but now this will break it:

```
# name = really long
```

18_regexp_lab/01.py
```diff
@@ -9,8 +9,9 @@
...
     for line in f:
-        if kuku.key in line:
-            if re.search('^\s+#', line):
-                continue
+        temp = line.split()
+        if temp[0].startswith('#'):
+            continue
+        elif temp[0] == kuku.key:
             res = re.search("=\s+(\w*)", line)
             print(res.group(1))
```
szabgab: This seems to select the proper key, but now you are just avoiding the regex. What you could have done earlier was using `*` to match any number of spaces in-front of the `#`.

```
if re.search('^\s*#', line):
   continue
```

The point if to play with regexes.

Anyway, there is still a problem. Try a prop file with

```
name = joe doe
```

Your code prints only 'joe' and leaves 'doe' out.

HananRosenthal: i have already uploaded a new (attitude) code (please check):

import argparse
import string
import re

parser = argparse.ArgumentParser(description='Returns a parameter that match a given key')
parser.add_argument('-p','--path', help='definitions file', required = True)
parser.add_argument('-k','--key', help='key', required = True)
kuku = parser.parse_args()

#Finally:
with open(kuku.path, 'r') as f:
    for line in f:
        prefix = re.match(r'\s_#', line)
        if prefix is not None:
            continue
        m = re.search(r'(\w+)\s_=\s_(._)', line)
        if m.group(1) == kuku.key and m is not None:
            print(m.group(1), '==>', m.group(2) )

HananRosenthal: בתרגול ביטויים רגולריים, הגשה, מופיעים: 01ת 02 03, אבל גם
: PROPS וגם INPUT
מה זה השניים האחרונים?

18_regexp_lab/01.py
```diff
@@ -1,28 +1,13 @@
...
-with open(kuku.path, 'r') as f:
-    for line in f:
-        temp = line.split()
-        if temp[0].startswith('#'):
-            continue
-        m = re.search(r'(\w+)\s*=\s*(\w*\s*\w*)', line)
+        m = re.search(r'(\w+)\s*=\s*(.*)', line)
         if m.group(1) == kuku.key and m is not None:
             print(m.group(1), '==>', m.group(2) )
-
```
szabgab: Much better, but we have lost the argparse part from the code so it is a bit hard to try it.
In any case a few issues are left:
- If there are spaces after the value they will be included in the value
- it would be enough to write `if prefix` for the condition in line 9.
- In `if m.group(1) == kuku.key and m is not None:` the conditions are in the wrong order. It would be much better to write `if m and m.group(1) == kuku.key`

18_regexp_lab/02.py
```diff
@@ -1,6 +1,25 @@
...
+
+def fromCamelCase(value):
+    return re.sub('([A-Z]+)', r'_\1', value).lower()
+
+
+test = '__hanan_rosenthal_The_second'
+res = toCamelClass(test)
+print( res)
+back = fromCamelCase(res)
+print(back)
```
szabgab, line 25: 👍 looks good

18_regexp_lab/03.py
```diff
@@ -1,15 +1,14 @@
...
+import re
+
+parser = argparse.ArgumentParser(description='')
+parser.add_argument('-p','--path', help='csv file', required = True)
+kuku = parser.parse_args()
+
+with open(kuku.path, 'r') as f:
+    reader = csv.reader(f)
+    for row in reader:
+        print ('{} {} {}'.format(row[1], row[0], row[2]) )
```
szabgab: 👍 
What if there are only 2 columns or more than 3?

18_regexp_lab/01.py
```diff
@@ -2,12 +2,17 @@
...
         prefix = re.match(r'\s*#', line)
-        if prefix is not None:
+        if prefix:
             continue
         m = re.search(r'(\w+)\s*=\s*(.*)', line)
-        if m.group(1) == kuku.key and m is not None:
-            print(m.group(1), '==>', m.group(2) )
+        if m and m.group(1) == kuku.key:
+            mm=m.group(2).strip() #remove whitespaces at the end of whatever string is
+            print(m.group(1), '==>', mm )
```
szabgab, line 18: 👍

18_regexp_lab/03.py
```diff
@@ -10,5 +9,12 @@
...
+    num_cols = len(next(reader))
+    f.seek(0 ,0)
+
     for row in reader:
-        print ('{} {} {}'.format(row[1], row[0], row[2]) )
+        if num_cols > 1:
+            row[0],row[1] = row[1],row[0]
+            print(row)
+        else:#print as is...
+            print(row)
```
szabgab: Seems good, but please don't rewind the file using seek. That should be used only for really low level libraries.

HananRosenthal: ok, but how else can i step back?

szabgab: Change the code that you don't need to step back.

18_regexp_lab/03.py
```diff
@@ -1,20 +1,14 @@
...
-    f.seek(0 ,0)
-
     for row in reader:
-        if num_cols > 1:
+        if len(row) > 1:
             row[0],row[1] = row[1],row[0]
-            print(row)
-        else:#print as is...
-            print(row)
+        print(row)
```
szabgab, line 14: 👍 Excellent

PR #176, opened 2016-07-19

zfigov:


18_regexp_lab/01.py
```diff
@@ -10,3 +10,17 @@
...
+import re
+
+(_,file1,key) = sys.argv
+
+
+lines_file1 = []
+
+with open(file1,"r") as fin1:
+    for line in fin1:
+        res = re.search(r'\b%s' % key,line)
```
ynonp, line 23: Using external input as regexp is considered bad practice: The input may contain characters that have special meaning in regexp, such as `.`, `^`, `$`, etc. Better to read all values to a dictionary and search that.

PR #178, opened 2016-07-19

nirna:


18_regexp_lab/02.py
```diff
@@ -1,6 +1,7 @@
-"""
-Write a function toCamelCase(text) that returns the text camel cased
-Write a function to_underscore(text) that returns the text with 
-underscores between words
-"""
-
+import re
+
+def toCamelCase(underscore_line):
```
ynonp, line 3: Function names in python are written in lowercase (same as variable), so use `to_camelcase`

18_regexp_lab/01.py
```diff
@@ -1,12 +1,11 @@
...
-"""
-
+import sys
+import re
+
+cfg_file_path = sys.argv[1]
+key_in = sys.argv[2]
+with open(cfg_file_path, 'r') as cfg_file:
+    for line in cfg_file:
+        res = re.search(r'^\s*'+key_in+r'\s*=\s*(.*)$', line)
```
ynonp, line 8: Using external input as regexp is considered bad practice: The input may contain characters that have special meaning in regexp, such as `.`, `^`, `$`, etc. Better to read all values to a dictionary and search that.

18_regexp_lab/03.py
```diff
@@ -1,15 +1,7 @@
...
-    Hampton,Witt,witthampton@zaphire.com
-    Grant,Morgan,morgangrant@lotron.com
-"""
+import re
+import sys
+
+input_csv = sys.argv[1]
+with open(input_csv) as fin:
+    for line in fin:
+        print re.sub(r'^([^,]*),([^,]*),', lambda m: m.group(2) + ',' + m.group(1) + ',', line.strip())
```
ynonp, line 7: 👍

PR #179, opened 2016-07-20

omerblecher:


20_classes_lab/01.py
```diff
@@ -1,6 +1,36 @@
...
+s.add(10, 20)
+t.add(50)
+s.add(30)
+
+# should print 60
+print s.total()
+
+# should print 50
+print t.total()
+
```
szabgab, line 36: 👍 looks good!

20_classes_lab/02.py
```diff
@@ -1,4 +1,18 @@
...
+    def __init__(self):
+        MyCounter.count += 1
+
+
+for _ in range(10):
+     c1 = MyCounter()
+
+# should print 10
+print MyCounter.count
+
```
szabgab, line 18: 👍

20_classes_lab/03.py
```diff
@@ -1,13 +1,72 @@
...
+darth.add_dependency(anakin)
+
+_all.add_dependency(luke, hansolo, leia, yoda, padme, anakin, obi, darth)
+_all.build()
+     
+
+
+
+  
+
```
szabgab, line 72: 👍 this looks excellent. 

A few comments:
- _index seems to be unnecessary. You could get the next number by calling `len(self._dependDict)` but:
- If you care about the order then it is usually better to have a list and not a dictionary

PR #180, opened 2016-07-20

konst1988:


20_classes_lab/01.py
```diff
@@ -4,3 +4,24 @@
 Complete the class so correct result is printed
 """
 
+class Summer(object):
+    def __init__(self):
+        self._sum = 0
+    def add(self, *nums):
```
ynonp, line 10: Python's style guide suggests to leave a blank line just before method definitions (and that also looks better). Here's the relevant paragraph from it:
https://www.python.org/dev/peps/pep-0008/#blank-lines

20_classes_lab/01.py
```diff
@@ -4,3 +4,24 @@
...
+        for num in nums:
```
ynonp, line 11: Next time better to use python's `sum` function - `self._sum += sum(nums)`

20_classes_lab/02.py
```diff
@@ -2,3 +2,14 @@
...
+     count = 0
+
+     def __init__(self):
+         MyCounter.count+=1
+
+for _ in range(10):
+     c1 = MyCounter()
+
+# should print 10
+print MyCounter.count
```
ynonp, line 15: Lovely!

20_classes_lab/03.py
```diff
@@ -11,3 +11,41 @@
...
+    def add_dependency(self, *depends):
+        self._depends.extend([x for x in depends])
+
+    def build(self):
+        self._isBuilt = True
+        for dep in self._depends:
+            if dep._isBuilt: continue
+            dep.build()
+            print dep._name
+
```
ynonp, line 30: 👍

PR #181, opened 2016-07-20

zfigov:


20_classes_lab/01.py
```diff
@@ -3,4 +3,26 @@
...
+class Summer(object):
+    def __init__(self):
+        self._current_sum = 0
+        
+    def add(self,*val):
+        for values in val:
+            self._current_sum += values
 
+    def total(self):
+        print self._current_sum
```
ynonp, line 15: Better to `return` than `print` here:
1. Program already has `print` when using the value, so you get double printing
2. In general - inside the class you never know what external code should decide to do with your results. Unless you have a good reason, prefer returning values from methods and functions

PR #183, opened 2016-07-20

dorsegal:


18_regexp_lab/01.py
```diff
@@ -9,4 +9,22 @@
...
+def getvaule(key, f):
+    #(?<=key = ).*
+    res = re.search(r"(?<=" + key + r' = ).*', f)
+    if res is not None:
+        return res.group(0)
+    else: return "key not found"
+
+with open(args.file, "r") as f:
+    print getvaule(args.key, f.read())
+
```
szabgab: Nice first version but it has a number of problems. e.g. if there is a line

```
# name = bob
```

and we are looking for the 'name' field?
or if the number of spaces around the = signe is not 1 and 1 then this regex won't match.

In general I'd probably first read in the whole file and fill a dictionary with the key-value pairs and then use that to look-up the requested key.

18_regexp_lab/01.py
```diff
@@ -11,20 +11,27 @@
...
+        res = re.search(r"^\W*" + args.key + r'\W*=\W*(.*)', line)
+        if res is not None:
+            dic[args.key].append(res.group(1))    
+            
+if dic.get(args.key) is not None:
+    for item in dic.get(args.key):
+        print item
+else:
+    print "key not found"
+
```
szabgab, line 37: It would be probably better to first read and parse the whole file into a dictionary and then use that to look up the expected field. I see you prepare for multiple entries with the same key. Nice, though usually that is disregarded in similar config files. The last one wins. (Though I think I'd add a warning message if I find a duplicate key.)

Lastly, instead of `if res is not None:` I'd just write `if res:` and instead of `if dic.get(args.key) is not None:` I'd write `if args.key in dic:` those are more Pythonic. Oh and I'd not use 'dic' as a variable name. That's just confusing.

PR #186, opened 2016-07-21

lahavz:


20_classes_lab/01.py
```diff
@@ -1,6 +1,32 @@
...
+"""
+Fill the code to print the result
+"""
+
+class Summer(object):
+    def __init__(self):
+        self._temp = []
+
+    def add(self, *args):
+        for x in args:
```
ynonp, line 10: Prefer list comprehensions - `self._temp += sum([...])`

20_classes_lab/01.py
```diff
@@ -1,6 +1,32 @@
...
-Complete the class so correct result is printed
-"""
-
+"""
+Fill the code to print the result
+"""
+
+class Summer(object):
+    def __init__(self):
+        self._temp = []
```
ynonp, line 7: Better to store just the sum. No need to keep all the items

20_classes_lab/01.py
```diff
@@ -1,6 +1,32 @@
...
+
+    def add(self, *args):
+        for x in args:
+            if (type(x) is int): self._temp.append(x)
+
+    def total(self):
+        sum = 0
```
ynonp, line 14: 1. Keeping just the sum as member is more efficient, since you don't need to calculate sum all over again every time `total` is called
2. `sum` is a bad variable name since it's already a builtin function in python

20_classes_lab/02.py
```diff
@@ -1,4 +1,18 @@
...
+    def __init__(self):
+        MyCounter.count += 1
+
+
+#Original Code
+for _ in range(10):
+     c1 = MyCounter()
+
+# should print 10
+print MyCounter.count
```
ynonp, line 18: 👍

20_classes_lab/03.py
```diff
@@ -1,13 +1,66 @@
...
-Implement widget so the following code works
-"""
-
+"""
+Fill the code to print the result
+"""
+
+class Widget(object):
+
+    widget_list = []
```
ynonp, line 7: Solution works but it's not recommended. Keeping a list of all the created widgets will keep them in memory even after nobody needs them. If you insist on using such a list it's better to use weak ref:
https://docs.python.org/2/library/weakref.html

However even with weakref it's not such a great idea - You still need to iterate the list and search for each widget to see if it was built.

A better pattern is to use normal instance variables here. 

Please modify the code not to use class variables

PR #187, opened 2016-07-21

Oransha:


18_regexp_lab/01.py
```diff
@@ -1,12 +1,12 @@
...
+file = sys.argv[1]
+user_input = raw_input("please enter your input:")
+
+
+with open(file,'r') as f:
+    for line in f:
+        if user_input in line:
+            print re.sub(r'(\w+)\s?=','', line)
+
+
```
szabgab, line 12: That's quite far from what the program should do. It should get the two parameters on the command line (and not ask the user), but more importantly if the user asks for 'name' this will return some value from
`# firstname = Foo`
but will not return from
`name     =   Foo`

18_regexp_lab/02.py
```diff
@@ -1,6 +1,29 @@
...
+        print text1, "not have big letter"
+    else:
+        lower = re.sub('([A-Z])',
+        lambda n: '_' + n.group(0).lower()
+        , text1) 
+        return lower
+
+
+
+print underline('hiThereHowAre')
```
szabgab, line 29: Nice.
One note: when you spread a function call on multiple lines, it is going to be more readable if the 2nd and further lines are indented. Even if that is not required by Python. Like this:

```
        upper = re.sub((r'_(\w)'),
            lambda m: m.group(0).upper(),
            text)
```

though I'd probably have written this in a single line. It is only 70 char wide.

Also there is no reason to call the internal variable in the second function 'res1'. That could be called 'res' too. Those two variables in the two functions are independent anyway.

PR #188, opened 2016-07-21

alonshk:


14_functions_lab/01.py
```diff
@@ -1,7 +1,18 @@
...
+	print sum
+
+def mymul(*args):
+	mul = 1
+	for i in args:
+		if type(i) == int:
+			mul *= i
+		else:
+			print str(i) + " is not an integer"
+	print mul
```
szabgab, line 18: 👍 though it should _ignore_ the non numeric values and not report them.

14_functions_lab/02.py
```diff
@@ -1,7 +1,8 @@
...
-"""
-
+
+
+def values(number, string):
+	if type(number) == int and type(string) == str:
+		return True
+	else:
+		raise Exception("First value should be number and second a string")
+
```
szabgab, line 8: 👍

14_functions_lab/03.py
```diff
@@ -1,5 +1,9 @@
...
-
+
+def mysum(*args):
+	total = 0
+	for i in args:
+		if isinstance (i, int):
+			total += (i / 10) % 10
+		else:
+			return False
+	print total
```
szabgab, line 9: 👍

PR #190, opened 2016-07-21

nirna:


20_classes_lab/01.py
```diff
@@ -1,6 +1,27 @@
...
-"""
-
+""" Class Summer: sums input numbers """
+class Summer(object):
+
+    def __init__(self):
+        self.sum = 0
+
+    def add(self, *nums):
+        for n in nums: self.sum += n
```
ynonp: Python already has a sum function - just for you: `self.sum += sum(nums)` (and that's also the reason you should use another name for the member variable - sum is a bit confusing)

20_classes_lab/03.py
```diff
@@ -1,13 +1,47 @@
...
+    def __init__(self,name):
+        self.name = name
+        self.dependencies = []
+
+    def add_dependency(self,*widgets):
+        for widget in widgets: self.dependencies.append(widget)
+
+    def build_internal(self,name_list):
+        if self.name in name_list: return []
+        for dependency in self.dependencies:
```
ynonp, line 14: What about nested dependencies ? When A depends on B which depends on C and so on. A flat loop won't catch everything

nirna, line 14: Nested dependencies are treated by the code since this is a recursion and not a flat loop.
It will fail in case loops exist (A-->B-->A) since I add the root's name to the list after the recursive call.
I can modify the code so root is added first, but this will break the print order unless I use one or another ugly workaround.

ynonp, line 14: Ah yes now I see I missed the recursion. Still you can greatly simplify this code by adding an is_built member field to the class (and this would also solve the loop).

When `build` is called on a widget, it would first check if it was already built

PR #191, opened 2016-07-22

shlomitTal:


18_regexp_lab/01.py
```diff
@@ -1,12 +1,14 @@
...
+import re
+if len(sys.argv) < 3: 
+    sys.exit("error - you didn't add any arguments!")
+
+regex= sys.argv[2] + '[ ]*=[ ]*(.*)'
+
+file = sys.argv[1]
+with open(file, "r") as f:
+    for line in f:
+        res = re.search(regex,line)
```
ynonp, line 12: Using external input as regexp is considered bad practice: The input may contain characters that have special meaning in regexp, such as `.`, `^`, `$`, etc. Better to read all values to a dictionary and search that.

ynonp, line 12: I'd read the entire file to a dictionary and then search there. It also yields a more robust solution for more complex situations

18_regexp_lab/02.py
```diff
@@ -1,6 +1,11 @@
-"""
-Write a function toCamelCase(text) that returns the text camel cased
-Write a function to_underscore(text) that returns the text with 
-underscores between words
-"""
-
+#commited by Shlomit
+import re
+
+def ToCamel(text):
```
ynonp, line 4: Python style is to use lowercase only names for functions and variables, so better use `to_camelcase` and `from_camelcase`

18_regexp_lab/03.py
```diff
@@ -1,15 +1,10 @@
...
-    Sargent,Shana,shanasargent@isoswitch.com
-    Hampton,Witt,witthampton@zaphire.com
-    Grant,Morgan,morgangrant@lotron.com
-"""
+#comited by Shlomit
+import re
+
+def changePlace(text):
+   regex= "(^[^,]+),([^,]),(.*)"
+   res = re.search(regex,text)
```
ynonp, line 6: Nice, but I think you can improve the code by using `re.sub`. Care to try?

PR #192, opened 2016-07-23

shlomitTal:


20_classes_lab/01.py
```diff
@@ -1,6 +1,15 @@
...
-"""
-
+#commited by Shlomit
+import sys
+class Summer(object):
+    def __init__(self):
+        self.sum = 0
+
+    def add(self, *list):
+        for n in list:
```
ynonp, line 8: Better to use python's `sum` function - `self.sum += sum(list)`

20_classes_lab/01.py
```diff
@@ -1,6 +1,15 @@
...
+            self.sum += int(n)
```
ynonp, line 9: Leave one blank line between method definitions in an object. 
See Python's style guide for more:
https://www.python.org/dev/peps/pep-0008/#blank-lines

PR #194, opened 2016-07-24

mashaGC:


18_regexp_lab/01.py
```diff
@@ -1,12 +1,26 @@
...
+        if m is not None:
+            key = m.group(1)
+            value = m.group(2)
+            kvDic[key] = value
+
+if kvDic.has_key(input_key):
+    print input_key," = ",
+    print kvDic.get(input_key)
+else:
+    print "key:'{}' doesn't exist".format(input_key)
```
szabgab, line 26: - What if the value of a key has spaces in it as in `name = Foo Bar`?
- `if m is not None:` is better written as `if m:`
- Instead of `if kvDic.has_key(input_key):` I'd write `if input_key in kvDic`

18_regexp_lab/02.py
```diff
@@ -1,6 +1,24 @@
...
+    if text == "":
+        break
+    toCamelCase(text)
+
+print 'toUnderscore'
+while True:
+    text = raw_input()
+    if text == "":
+        break
+    toUnderscore(text)
```
szabgab, line 24: 👍

18_regexp_lab/03.py
```diff
@@ -1,15 +1,8 @@
...
-    Grant,Morgan,morgangrant@lotron.com
-"""
+_author_ = 'mariag'
+
+import re
+lines = [line.rstrip('\n') for line in open('input.csv')]
+
+for line in lines:
+    m = re.search('^([^,]*),([^,]*),(.*)', line)
+    print "{},{},{}".format(m.group(2),m.group(1),m.group(3))
```
szabgab, line 8: 👍

PR #195, opened 2016-07-24

dovron:


20_classes_lab/01.py
```diff
@@ -1,6 +1,33 @@
...
+
+"""
+
+class Summer(object):
+
+    def __init__(self):
+        self.item = 0
+
+    def add(self, *num):
+        self.item += sum([n for n in num if type(n) == int])
```
ynonp, line 17: Personally I'd skip the test for `int` here: If `add` is called with non-numeric arguments it probably was by mistake. By breaking you force your users to face their mistake and fix their code. By ignoring the non-numeric values such mistakes could stay in code longer and cause hidden subtle mistakes later on.

PR #196, opened 2016-07-24

eladagam:


18_regexp_lab/01.py
```diff
@@ -1,12 +1,28 @@
...
+key = args['key']
+
+
+with open(conf,"r") as fin:
+    for line in fin:
+        res = re.search(key+r'.*=.*(\d)+', line)
+        if res is not None:
+            print res.group(1)
+
+
```
szabgab, line 28: It is better to first read and parse the whole file into a dictionary and then look up the key.
Besides this will match the commented out`# answer = 42` when searching for _answer_ but won't match `name = Foo Bar` when searching for _name_.

18_regexp_lab/02.py
```diff
@@ -1,6 +1,14 @@
...
+
+import re
+import argparse
+import os
+
+def toCamelCase(text):
+    str =  re.sub(r'_[a-z]', lambda m: m.group(0).upper(), text)
+    print re.sub(r'_','' , str)
+
+toCamelCase("no_more_shall_we_part")
```
szabgab, line 14: 👍

18_regexp_lab/03.py
```diff
@@ -1,15 +1,26 @@
...
+source = os.path.join(workingDir, args['source'])
+
+
+
+
+with open(source,"r") as fin:
+    for line in fin:
+        print  re.sub(r'^(\w+),(\w+),', r'\2,\1,', line)        
+        #I can't understand why the below did not work!!!
+        #print  re.sub(r'^(\w+),(\w+),', lambda m:m.group(1)+","+m.group(0), line)
```
szabgab, line 26: - You don't need to prepend the current working directory. A plain relative path will work.
- re: your question: Maybe if you describe what happened it would be easier to understand.

eladagam, line 26: re.sub(r'^(\w+),(\w+),', lambda m:m.group(1)+","+m.group(0), line) -
did not work
It just added another instance of the first column instead of
replacing the order

szabgab, line 26: group(0) ?   didn't you want to use group 1 and 2 there?

eladagam, line 26: Wow - that's true!!

PR #197, opened 2016-07-24

alonshk:


12_listcomprehension_lab/01.py
```diff
@@ -1,8 +1 @@
-"""
-Use range() and list comprehension to get
-the list of all lowercase english letters
-Hint: look for chr() and ord()
-"""
-
-
-
+map(chr, range(ord('a'), ord('z')+1))
```
szabgab, line 1: That's a solution, but that's not a list comprehension.

12_listcomprehension_lab/02.py
```diff
@@ -1,6 +1,3 @@
-"""
-Write a python program that takes two words
-as sys.argv and prints only the letters
-common to both
-"""
-
+
+lst = map(chr, range(ord('a'), ord('z')+1))
+print [x + y + z for x in lst for y in lst for z in lst]
```
szabgab, line 3: 👍

12_listcomprehension_lab/01.py
alonshk, line 1: ok
so what is it?
looks similar to reduce function..

2016-07-24 22:53 GMT+03:00 Gabor Szabo notifications@github.com:

szabgab, line 1: Yes, it is a reduce function and it is a good solution in general. It is just not called a list comprehension which would have the form of `[ ... ]`.

PR #198, opened 2016-07-24
ToCode Solution: תרגול קבצים

idisis:


08_files_lab/01.py
```diff
@@ -1,6 +1,19 @@
-"""
-Write a program that takes 2 file names
-and appends the second file's contents to
-the end of the first one
-"""
-
+import argparse
```
ynonp: You accidentally submitted the same solution twice

ynonp, line 1: You accidentally submitted the solution to the wrong exercise here. Should take 2 file named and append the second file's contents to the end of the first one

08_files_lab/02.py
```diff
@@ -1,7 +1,18 @@
...
+
+with open(args.a, 'r') as a:
+  with open(args.b, 'r') as b:
+    with open(args.o, 'w') as o:
+        for line_tuple in izip_longest(a, b):
+            for line in line_tuple:
+                if line != None:
+                   o.write(line)
+                   if line[-1] != '\n':
+                     print >> o, ""
```
ynonp, line 18: 👍

08_files_lab/01.py
idisis, line 1: Sorry, I just forgot to update the description of the script in the argparse...

08_files_lab/01.py
```diff
@@ -16,4 +16,4 @@
         for line in b:
             o.write(line)
```
ynonp, line 17: Cool. Another idea - you can replace the if with a single: `print >>o, line.rstrip()`, so you cover both cases

PR #199, opened 2016-07-24

dovron:


20_classes_lab/03.py
```diff
@@ -1,13 +1,49 @@
...
+        self.depends = []
+        self.complete = False
+
+    def add_dependency(self, *Widget):
+        self.depends.extend(Widget)
+
+    def build(self):
+        self.complete = True
+        for dep in self.depends:
+            if not dep.complete:
```
ynonp: Fantastic! One style comment: It's usually nicer to `continue` on wrong input than to perform on the right ones. So prefer:

```
for dep in self.depends:
  if dep.complete: continue

  dep.build()
  print dep.name
```

ynonp:
Great work. Closing.

PR #201, opened 2016-07-25

tomag:


20_classes_lab/01.py
```diff
@@ -4,3 +4,23 @@
 Complete the class so correct result is printed
 """
 
+class Summer(object):
+    def __init__(self):
+        self.sum = 0        
+
+    def add(self, *values):
+        for value in values:
```
ynonp, line 12: Better to use python's sum: `self.sum += sum(values)`. Also `sum` is a confusing name for member field (because of the builtin function)

20_classes_lab/03.py
```diff
@@ -11,3 +11,40 @@
...
+    def __init__(self, name):
+        self._name = name
+        self._dependencies = []
+        self._is_built = False
+
+    def add_dependency(self, *widgets):
+        self._dependencies.extend(widgets)
+
+    def build(self):
+        if (self._is_built): return
```
ynonp, line 25: Nice! Also you don't need parenthesis `if self._is_built: return` looks better

PR #204, opened 2016-07-25

idisis:


10_datastructures_lab/01.py
```diff
@@ -1,10 +1,20 @@
...
-"""
+from sys import argv
+
+hostnames = set(argv[1:])
+hosts = {}
+
+# parse hosts file:
+with open('hosts', 'r') as hostsFile:
+    for line in hostsFile:
+        (host, address) = line.split('=', 2)
```
ynonp, line 9: No need for parenthesis here - python prefers: `host, address = line.split('=', 2)`

10_datastructures_lab/01.py
```diff
@@ -1,10 +1,20 @@
...
+        if address[-1] == '\n':
```
ynonp, line 10: Strings have `rstrip()` function which deletes all newline characters and is preferred

10_datastructures_lab/01.py
```diff
@@ -1,10 +1,20 @@
...
+            address = address[:-1]
+        hosts[host] = address
+
+for name in hostnames:
+    if  hosts.has_key(name):
```
ynonp, line 15: Better style is to use `if name in hosts:`

PR #205, opened 2016-07-25

tomag:


25_exceptions_lab/01.py
```diff
@@ -4,4 +4,22 @@
...
 a warning and keep reading values
 """
+from math import sqrt
 
+def mySqrt(val):
+    if not (type(val) == int):
+        raise TypeError("Value is not integer")
+    elif (val < 0):
+        raise ValueError("Value must be greater than 0")
+    return sqrt(val)
```
ynonp, line 14: Python's sqrt already raises exceptions, so no need to wrap it in your own function. Just catch the exception and handle them correctly

25_exceptions_lab/02.py
```diff
@@ -5,3 +5,17 @@
 Alert the user politely if there was any problem opening the file
 """
 
+import os.path
+import sys
+
+
+fileName = sys.argv[1]
+
+if not (os.path.isfile(fileName)):
```
ynonp, line 14: It is generally preferred to perform an action AND catch the exception than to check `os.path.isfile` first:
1. There are tons of reasons why opening a file can fail (no permissions, IO errors, etc.). Verifying something is a file doesn't cover all the cases.
2. Calling `os.path.isfile` is an extra unneeded operation (which is a waste).
3. When `os.path.isfile` fails you still raise an exception, so you'll still need to add an except block to the `with open` command

25_exceptions_lab/03.py
```diff
@@ -1,4 +1,36 @@
...
+        self._message = message
+    
+    def __str__(self):
+        return self._message
+
+class ImageFile(object):
+    validExtension = "png"
+    
+    def __init__(self, fileName):
+        result = re.search("[a-z,0-9]*.(?P<ext>[a-z,0-9]+)", fileName)
```
ynonp, line 18: Python has an `os.path.splitext` function that does a better job than writing your own regexp

25_exceptions_lab/03.py
```diff
@@ -1,4 +1,36 @@
...
 
+class InvalidImageExt(Exception):
+    def __init__(self, message):
+        self._message = message
+    
+    def __str__(self):
+        return self._message
+
+class ImageFile(object):
+    validExtension = "png"
```
ynonp, line 15: I'd use a list here, to allow future code to add more extensions

PR #206, opened 2016-07-25

konst1988:


25_exceptions_lab/01.py
```diff
@@ -4,4 +4,16 @@
 If value is negative or not a number show 
 a warning and keep reading values
 """
+import sys
+import math
 
+while True:
+    number = raw_input("Enter number:")
+    if number == 0: sys.exit()
```
ynonp, line 12: why?

25_exceptions_lab/01.py
```diff
@@ -4,4 +4,16 @@
...
+    try:
+        result = math.sqrt(int(number))
+        print "SQRT of {} is {}".format(number, result)
+    except ValueError as e:
+        print "Invalid number for sqrt. Error: {}".format(e.message)
+    except Exception as e:
+        print "Unknown Error occurred: {}".format(e.message)
```
ynonp, line 19: Great!

25_exceptions_lab/02.py
```diff
@@ -4,4 +4,14 @@
 
 Alert the user politely if there was any problem opening the file
 """
+import sys
 
+file = sys.argv[1]
+
+try:
+    with open(file, "rb") as fin:
```
ynonp, line 12: When opening the file as "binary" you should use the binary read operations (`f.read(bufsize)`). 

When reading lines open the file as text (omitting the "b")

25_exceptions_lab/03.py
```diff
@@ -1,4 +1,35 @@
...
+class InvalidImageExt(Exception):
+
+    def __init__(self, file):
+        super(InvalidImageExt, self).__init__("File '{}' has an invalid image extension".format(file))
+
+
+class ImageFile(object):
+
+    def __init__(self, file):
+        if not re.search(r'[.]gif|png|jpg|jpeg$', file):
```
ynonp, line 18: Python has a `os.path.splitext` function which does a better job than writing your own regexp.

Regarding the regexp notice you need parenthesis around extensions because $ only refers to the last alternative (for example a.giff matches your regex -  try `re.search(r'[.]gif|png|jpg|jpeg$', '.giff')`

PR #208, opened 2016-07-25

Oransha:


Oransha:
I don't know how to receive input with 2 arguments like s.add(10, 20)

20_classes_lab/01.py
```diff
@@ -1,6 +1,22 @@
...
+s.add(10, 20)
+t.add(50)
+s.add(30)
+
+# should print 60
+print s.total()
+
+# should print 50
+print t.total()
+
```
szabgab, line 22: Keeping all the numbers and calculating the total only on request is an interesting solution.
Also using list comprehension 🍪 
👍

szabgab, line 22: For accepting 2 or more values you might want to take a look at this SO question:
http://stackoverflow.com/questions/36901/what-does-double-star-and-star-do-for-python-parameters

20_classes_lab/02.py
```diff
@@ -1,4 +1,11 @@
...
+    count = 0
+
+    def __init__(self):
+        MyCounter.count += 1
+
+for _ in range(10): 
+    c1 = MyCounter()
+
+# should print 10
+print MyCounter.count
```
szabgab, line 11: 👍

PR #210, opened 2016-07-26

eladagam:


25_exceptions_lab/01.py
```diff
@@ -1,7 +1,20 @@
...
+        num = float(numStr)
+    except ValueError:
+        print ("%s is not a number\n" % numStr)
+        exit(0)
+    
+    
+    if num < 0:
+        raise ValueError("Expected parameter >= 0, Got: %d" % num)
+
+    print math.sqrt(num)
```
szabgab: Now try to do it the other way around. Let math.sqrt throw an exception, catch it and then print an appropriate warning based on the type of the exception you got.

25_exceptions_lab/02.py
```diff
@@ -1,7 +1,14 @@
...
+import os
+import sys 
+
+
+if not os.path.isfile(sys.argv[1]):
+    raise ValueError("%s does not exist\n" % fin)
+
+with open(sys.argv[1],"r") as fin:
+    print sum(1 for _ in fin)
+
```
szabgab, line 18: I think "politely" here means without an exception, but also what happens if the file exists but you don't have rights to read it?

I'd not use the 'sum' function here, instead I'd just increment a variable for every line because I think the above code will generate a list of as many 1s as there are lines which means it might be a really long list.

25_exceptions_lab/01.py
```diff
@@ -8,13 +8,19 @@
...
     
     
-    if num < 0:
-        raise ValueError("Expected parameter >= 0, Got: %d" % num)
 
-    print math.sqrt(num)
+    
+    
+
+
```
szabgab, line 26: 👍 comparing the error string is usually not a good idea as that string might change in the future or it might depend on the current locale, but in this case you don't have a choice if you really want to give your own message instead of the e.message

25_exceptions_lab/02.py
```diff
@@ -1,14 +1,19 @@
...
-with open(sys.argv[1],"r") as fin:
-    print sum(1 for _ in fin)
+    with open(sys.argv[1],"r") as fin:
+        for _ in fin:
+            sum=sum+1
+except Exception as e:
+    print  ("Sorry, file %s not found" % sys.argv[1])
+    exit(0)
 
+print sum
```
szabgab, line 19: 👍 
Three minor issues:
- I'd write sum += 1
- If you run the script without any parameter it throws an exception.
- Better use .format instead of % for string formatting.

25_exceptions_lab/03.py
```diff
@@ -1,4 +1,36 @@
...
+        try:
+            img = ImageFile("file.png")
+        except InvalidImageExt:
+            self.fail("png should be a valid file extension")
+
+    def test_bad_ext(self):
+        with self.assertRaises(InvalidImageExt):
+            img = ImageFile("file.mp3")
+
+unittest.main()
```
szabgab, line 36: 👍

PR #212, opened 2016-07-27

HananRosenthal:


20_classes_lab/01.py
```diff
@@ -1,6 +1,28 @@
...
+
+s.add(10, 20)
+t.add(50)
+s.add(30)
+
+# should print 60
+print(s.total())
+
+# should print 50
+print(t.total())
```
szabgab, line 28: Nice, but you don't need the list comprehension. You could write `self.val += sum(num)`

20_classes_lab/02.py
```diff
@@ -1,4 +1,17 @@
...
+    @classmethod
+    def count(self):
+        return sum(MyCounter.counter)
+
+
+for _ in range(10):
+    c1 = MyCounter()
+
+# should print 10
+print(MyCounter.count())
```
szabgab: I don't see the value in keeping a list of 1s and then sum()-ing them. Why not just keep a counter variable?

HananRosenthal: how can i define counter as int without initializing it? in that case each construction will overides its value. How can i define an int variable in the class?

szabgab: You can define it the same way you defined the counter list. However, inside the methods you need to prefix it with the name of the class as in `MyCounter.counter`.

20_classes_lab/01.py
```diff
@@ -7,7 +7,7 @@ def __init__(self):
         self.val = 0
 
     def add(self, *num):
-        self.val += sum([i for i in num])
+        self.val += sum(num)
 
     def total(self):
         return self.val
```
szabgab, line 13: 👍

20_classes_lab/02.py
```diff
@@ -1,17 +1,16 @@
...
-        return sum(MyCounter.counter)
-
+        return MyCounter.counter
 
 for _ in range(10):
     c1 = MyCounter()
 
 # should print 10
-print(MyCounter.count())
+print(MyCounter.count(object))
```
szabgab, line 16: The count method does not need 'self' and if you declare it without any parameter then you don't need to pass 'object' to it which is not the instance object in any case.

PR #214, opened 2016-07-27

omerblecher:


szabgab:
Looks good. For now I found this exception. I think when exiting.

```
Exception in Tkinter callback
Traceback (most recent call last):
  File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/lib-tk/Tkinter.py", line 1532, in __call__
    return self.func(*args)
  File "main.py", line 59, in start
    p2 = player.player(inputO.value, "O")
AttributeError: 'inputWindow' object has no attribute 'value'
```

omerblecher:
I fixed it - I haven't handled a case when the user doesn't enter the player's name or when he clicks on the Close X sign

szabgab:
Nice. Now try to run the unit-tests with `python -m unittest discover`

omerblecher:
I can run it, but I haven't seen that I can submit the unit test file

szabgab:
No need to submit the unittest file. No need to even change the unittest file.

omerblecher:
Need to change the unittest file since there are methods that doesn't exist in my solution and the data structure for the board is different in my implementation :)

szabgab:
I see. Then either change your solution, or change the tests to pass but don't worry about submitting them.

PR #216, opened 2016-07-27

tomag:


27_tictactoe/main.py
```diff
@@ -0,0 +1,70 @@
...
+
+    playGame()
+
+    print
+    print "Another game? (Y|N)",
+    decision = raw_input()
+    if decision.lower() == 'n':
+        break
+
+print "Bye!"
```
ynonp, line 70: General Comments: 
1.  Unit tests fail (You can try to run them with `python -m unittest discover` from command line).
2.  Error handling: Wrong user input is part-of-the-deal when writing this sort of game. It should handle invalid inputs better and allow users to keep playing
3. 👍  I loved the custom game feature and support for any board size! That's cool
4. Prefer lowercase names for functions and variables

ynonp, line 70: All in all great work in the project and in the course!

PR #217, opened 2016-07-27

moshekenan:


25_exceptions_lab/01.py
```diff
@@ -1,7 +1,17 @@
...
-and for each number prints its square root
-If value is negative or not a number show 
-a warning and keep reading values
-"""
-
+import math
+def sqrt(n):
+    if n < 0:
+        raise ValueError("Expected parameter >= 0, Got: %d" % n)
+    return math.sqrt(n)
```
ynonp, line 5: No need to write your own `sqrt` function - python already has one and it's already raising the correct exceptions.

25_exceptions_lab/02.py
```diff
@@ -1,7 +1,15 @@
...
-"""
-
+import sys
+import os
+
+src = sys.argv[1]
+
+line_count = 0
+
+if ( not os.path.isfile(src)):
```
ynonp, line 8: No need to check if `src` is a file:
1. Open could fail for a million other reasons, so it's not that productive anyways
2. If everything's ok, you just wasted time checking something that's true

A better (more pythonic) way is to use `except` after `with open`, handling exceptions if they happen

25_exceptions_lab/03.py
```diff
@@ -1,4 +1,32 @@
...
+import shutil
+
+class InvalidImageExt(Exception):
+    def __init__(self, msg):
+        super(InvalidImageExt, self).__init__(msg)
+
+class ImageFile:
+    def __init__(self, name):
+        self.path = name
+        file_extension = name.split(".")[-1]
```
ynonp, line 13: Better to use `os.path.splitext` to extract name and extension from a path

25_exceptions_lab/03.py
```diff
@@ -1,4 +1,32 @@
...
+        print file_extension
+        if file_extension != 'png':
+            raise InvalidImageExt("png should be a valid file extension") 
+
+        open(self.path, 'a').close()
```
ynonp, line 18: Why do you need to open and close the file?

PR #218, opened 2016-07-27

eladkah:


25_exceptions_lab/01.py
```diff
@@ -1,7 +1,31 @@
...
+    except ValueError:
+        raise ValueError('Input \'%s\' is not a number' % string)
+
+
+def sqrt(n):
+    if type(n) == str:
+        n = try_parse_float(n)
+    if n < 0:
+        raise ValueError('Expected positive number, got %f' % n)
+    return math.sqrt(n)
```
ynonp, line 21: Hi Elad,
No need to add complexity where it's not required - 

Better (more pythonic) approach is to use python's built-in math.sqrt function directly, because it already raises the correct exceptions. Don't try to parse, convert or check - just do what you need to and add an except block to handle failures

PR #219, opened 2016-07-27

nirna:


25_exceptions_lab/01.py
```diff
@@ -4,4 +4,12 @@
 If value is negative or not a number show 
 a warning and keep reading values
 """
+import math
 
+while True:
+    try:
+        n = float(raw_input("Enter a nmber: "))
+        print "The square root of %f is %f" %(n,math.sqrt(n))
+
```
ynonp, line 13: 👍

25_exceptions_lab/02.py
```diff
@@ -4,4 +4,24 @@
...
+
+# Define input arguments
+parser = argparse.ArgumentParser()
+parser.add_argument("file_path", type=str, help="The file path to count lines of.")
+
+# Read input arguments
+args = parser.parse_args()
+
+try:
+    if not os.path.exists(args.file_path):
```
ynonp, line 18: No need to check for path existence. If you got the wrong path open will raise an exception. Prefer to add an except block to open

25_exceptions_lab/03.py
```diff
@@ -1,4 +1,31 @@
...
+import unittest
+
+class InvalidImageExt(Exception):
+    def __init__(self):
+        self.message = "Not an image extension"
+
+class ImageFile(object):
+    def __init__(self,file_name):
+        file_name = file_name.strip()
+        ext = re.search(r'\.(\w+)$', file_name)
```
ynonp, line 14: `os.path.splitetx` is a better way to get a file's extension than using a regexp (usually faster and more accurate)

25_exceptions_lab/03.py
```diff
@@ -1,4 +1,31 @@
...
+        if ext.group(1) in {'jpg','bmp','png','gif'}:
+            open(file_name,'w').close
```
ynonp, line 16: Why open?

PR #220, opened 2016-07-28

zfigov:


25_exceptions_lab/02.py
```diff
@@ -5,3 +5,16 @@
...
 """
 
+import sys
+import msvcrt
+
+(_,file1) = sys.argv
+
+try :
+    num_lines = sum(1 for line in open(file1))
+except Exception as e:
```
ynonp, line 15: Better to also print the exception itself: It may contain valuable information about what went wrong

25_exceptions_lab/03.py
```diff
@@ -2,3 +2,29 @@
...
 """
 
+import unittest
+
+class ImageFile:
+    def __init__(self,img_name):
+        # test valid file extention 
+        if img_name.lower().endswith('.png'):       
+            self.img_name = img_name
+            return 1
```
ynonp, line 12: No need to return from `__init__`

25_exceptions_lab/03.py
```diff
@@ -2,3 +2,29 @@
...
+        else:
```
ynonp, line 13: A better structure is:

```
if invalid_extension: raise InvalidImageExt(...)

# correct code written here with correct indentation
```

ynonp:
Great Solution. Closing.

PR #221, opened 2016-07-29

nirna:


27_tictactoe/main.py
```diff
@@ -0,0 +1,46 @@
...
+    if tto.tie(): print "Game tied!!"
+    else: print "Winner is %s - well done!!\n" %(tto.winner.name)
+    print "------- GAME OVER --------\n"
+
+    # Continue?
+    if raw_input("Would you like to play again (y/n)? ").lower() == 'n': 
+        print "\nGood bye... Hope you enjoyed!!\n"
+        break
+    if raw_input("Same players (y/n)? ").lower() == 'n': 
+        i = 0
```
ynonp, line 46: General Comments:
1. Unit tests don't pass (run with `python -m unittest discover`)
2. Gameplay is fun! Error handling works well
3. Extra Points for the docstring documentation (however docstring for methods should appear as the first lines inside the method, not before it)

All in all very good solution.

27_tictactoe/lib/board.py
```diff
@@ -0,0 +1,79 @@
...
+class Game:
+    def __init__(self):
+        self.reset_board()
+        self.players = [None,None]
+        self.current_player = None
+        self.winner = None
+        self.moves_counter = 0
+    
+    def reset_board(self):
+        self.data = [ [None,None,None] ,
```
ynonp, line 29: Python has \* operator for lists, so this also works: `self.data = [[None] * 3 for x in range(3)]`

27_tictactoe/lib/player.py
```diff
@@ -0,0 +1,11 @@
+
+class Player:
+
+    def __init__(self, name, marker):
+        if type(name)!=str: raise ValueError('Player name must be a string')
```
ynonp, line 5: Type checking is not necessary here: You can always try to convert to string, and if it fails python automatically raises an exception

27_tictactoe/main.py
nirna, line 46: All tests passed for me last night and still pass today…
Please advise.

[cid:image001.png@01D1E978.D48C8020]

ynonp, line 46: Interesting. Did you change anything in the test files themselves ? Because they're not part of the exercise and you can't submit them

nirna, line 46: I changed one of the tests since it was defective and added import for the Player class which was missing.
See attached.

ynonp, line 46: Problem was not in Player class but in `tie` detection. In the test `play` method is expected to get (col, row), and in your game it takes (row, col). That's why the play sequence in the test does not product a tie and the test fails.

This is the sequence that according to the test was to product a tie, but in your game produces victory for the first player:

 28         self.g.play(0,0)
 29         self.g.play(0,1)
 30         self.g.play(0,2)
 31         self.g.play(1,1)
 32         self.g.play(1,0)
 33         self.g.play(1,2)
 34         self.g.play(2,1)
 35         self.g.play(2,2)
 36         self.g.play(2,0)

That's not too bad, just solving in a different way.

nirna, line 46: I still believe that there was an issue with the original test since taking (col,row) instead of (row,col) should have the same effect as transposing the board which should make no difference on the result.
The sequence below generates:
X O X
O X O
X X O

Which means X wins.

ynonp, line 46: Oh my you're right... It was a mistake in the test. Thanks for the attention!
(the last two should have been the other way around).

This is the new test file, and now all your tests pass:
https://github.com/ynonp/python-examples-verint-2016-07/blob/master/27_tictactoe/tests/test_board.py

27_tictactoe/lib/board.py
nirna, line 29: This short writing leads to a bug in the program since this structure behaves differently.
Do you happen to know the reason?

I opened an issue for it: https://github.com/ynonp/python-examples-verint-2016-07/issues/222#issue-168286861

Thanks, Nir

ynonp, line 29: Fascinating. Updated my comment and wrote a detailed response in the issue.

PR #224, opened 2016-07-30

alonshk:


alonshk:
if the value of a key has spaces in it as in name = Foo Bar i will get just Foo. What should i do to resolve this ?

szabgab:
e.g.:

`````` re.search(r'(\w+)\s*=\s*(.+)')```
and then you can remove the trailing spaces from the value, but I might use split with '=' instead of this regex.

BTW This solution will also catch items that were commented out:
```# name = value```
that should be avoided.
``````

PR #226, opened 2016-07-30

shlomitTal:


25_exceptions_lab/02.py
```diff
@@ -1,7 +1,11 @@
...
+if len(sys.argv) < 2 :
+    raise BaseException("dont enter a file as argument!")
+try:
+    count = 0
+    with open(sys.argv[1], "r") as f:
+        for line in f:
+            count +=1
+    print count
+except Exception as e:
+    print "sory, file %s, not exist"% sys.argv[1]
```
ynonp, line 11: 👍

25_exceptions_lab/01.py
```diff
@@ -1,7 +1,22 @@
...
-If value is negative or not a number show 
-a warning and keep reading values
-"""
-
+#commited by Shlomit
+
+import numbers
+import sys
+import math
+def sqrt(n):
```
ynonp, line 6: Don't do that- Python already has a sqrt function that already raises exceptions. Just use existing and handle the exception in the bottom loop

